PR #13, opened 2020-07-27
Creates an html page that allows users to submit a form to create a destination for scavenger hunts

rysimone:


destinationCreation.html
```diff
@@ -0,0 +1,55 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
```
lilymzhou, line 3: The line spacing looks a little off here -- it seems like there's an extra indent.

destinationCreation.html
```diff
@@ -0,0 +1,55 @@
...
+            <label for="tourist-input">Tourist</label>
+            <input type="checkbox" name="family" id="family-input">
+            <label for="family-input">Family</label>
+            <button type="submit">Submit</button>
+          </form>
+        </div>
+        <p>* = a required field</p>
+      </div>
+    </body>
+</html>
```
lilymzhou, line 55: The HTML Presubmit requires an empty line at the end of html/css files.

rysimone: ok good to know. Thanks

destinationCreation.css
```diff
@@ -0,0 +1,5 @@
+#content {
```
tjgoog, line 1:
```suggestion
/* TODO: Move common styles to a core css file. */
#content {
```

Because it will be good to use consistent page width, and probably other styles as time goes on, across all pages.

destinationCreation.html
```diff
@@ -0,0 +1,55 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+      <meta charset="UTF-8"/>
+      <title>Destination Creation</title>
+      <link rel="stylesheet" href="destinationCreation.css"/>
+      <script src="destinationCreation.js"></script>
```
tjgoog, line 7: Does this file exist yet? Maybe it does and github's diff view is just hiding it. But if not, could you either create the file as just empty or a blank line, or remove this `script` tag?

destinationCreation.html
```diff
@@ -0,0 +1,55 @@
...
+    </head>
+    <header>
```
tjgoog, line 9: I think this should be removed since it's empty. Also I'm not sure if it's permitted outside of the `<body>`. But if some automated tool told you to put it in, that's fine.

destinationCreation.html
```diff
@@ -0,0 +1,55 @@
...
+    </header>
+    <body>
+      <div id="content">
+        <h1>Destination Creation Form</h1>
+        <div id="form-container">
```
tjgoog, line 14: I think this outer `<div>` can be removed

destinationCreation.html
```diff
@@ -0,0 +1,55 @@
...
+          <form action="/data" method="post">
+            <label for="name-input">Enter the name of the destination: *</label>
+            <textarea name="name" id="name-input" required></textarea>
+            <p>Enter the latitude and longitude of the destination:</p>
+            <label for="lat-input">Latitude: *</label>
+            <textarea name="latitude" id="lat-input" required></textarea>
+            <label for="long-input">Longitude: *</label>
+            <textarea name="longitude" id="long-input" required></textarea>
```
tjgoog, line 22: I recomment `lng` rather than `long` as the abbreviation. `long` may lead to confusion with the `long` integer type.

destinationCreation.html
```diff
@@ -0,0 +1,55 @@
...
+    <head>
+      <meta charset="UTF-8"/>
+      <title>Destination Creation</title>
+      <link rel="stylesheet" href="destinationCreation.css"/>
+      <script src="destinationCreation.js"></script>
+    </head>
+    <header>
+    </header>
+    <body>
+      <div id="content">
```
megankj, line 12: does this need to be an ID?

destinationCreation.html
```diff
@@ -0,0 +1,55 @@
...
+        <h1>Destination Creation Form</h1>
+        <div id="form-container">
+          <form action="/data" method="post">
```
megankj, line 15: /data will be the path for the POST request, so I think giving it a more descriptive name like /create-destination might make more sense

destinationCreation.html
```diff
@@ -0,0 +1,55 @@
...
+            <label for="name-input">Enter the name of the destination: *</label>
```
megankj, line 16: Instead of adding a "*" to each required input, you can do this with a class like this:

```
.required {
  color: red;
}
.required::after {
  content: "*";
}
```
You can make a required class and add it to the labels that are required.  This will let you style all the required labels the same way and add the little asterisk at the end!
https://www.w3schools.com/cssref/sel_after.asp

PR #22, opened 2020-08-03
Add enhancements to existing prototype features in the go feature.

lilymzhou:
- Add style (style.css)
- Hide hint and proceed buttons when they're unneeded, and show only once they're needed.
- User can exit the hunt by clicking an exit button. A window pops up confirming that they want to exit.
- Add additional messages to display when appropriate: a message appears to tell the user if their guess is correct or incorrect; a summary of all destinations found is displayed after the user finishes the hunt.

Start:
PR1_P1

PR1_P2

After inputting the correct destination name:
PR1_P4

After completing the hunt:
PR1_P5


src/main/java/com/google/sps/servlets/GoDataServlet.java
```diff
@@ -42,7 +45,15 @@
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     String indexStr = request.getParameter(INDEX_PARAMETER);
-    index = Integer.parseInt(indexStr);
+    try {
+      index = Integer.parseInt(indexStr);
+    } catch (Exception e) {
+    }
+
+    name = request.getParameter(NAME_PARAMETER);
```
megankj, line 53: It's not clear that name is a guess, especially considering the new argument to the ScavengerHunt constructor is called "guess".

```suggestion
    nameGuess = request.getParameter(NAME_PARAMETER);
```

src/main/webapp/go.js
```diff
@@ -18,22 +18,48 @@ const PROCEED_BUTTON = 'proceed-button';
...
-let destIndex;
-const destArr = [];
+let destIndex; // Marks the destination that the user currently needs to find.
+let hintIndex = 0; // Marks the hint that the user will see next.
+const huntArr = []; // Stores scavenger hunt data retrieved from the server.
+
+/**
+ * Represents a destination on the scavenger hunt.
+ */
+class Destination {
```
megankj, line 44: create a new file for this new class, destination.js

src/main/webapp/go.js
```diff
@@ -44,49 +70,84 @@ window.onload = getHunt();
...
 /**
- * Retrieves the string the user submitted as the destination name.
- * This data is fetched from the server because entity extraction
- * will be used in the MVP.
- * @return {boolean} Whether or not the user entered the correct name.
+ * Add a marker to the map at the specified location.
+ * @param {Double} destLat: latitude of location.
+ * @param {Double} destLng: longitude of location.
+ * @param {String} destName: Name of location.
+ * This function is implemented in another PR.
```
megankj, line 113:
```suggestion
 * TODO: Implement this function.
```

src/main/webapp/go.js
```diff
@@ -44,49 +70,84 @@ window.onload = getHunt();
...
-    if (mssg === destArr[destIndex]) { // The user entered the correct name.
-      return true;
-    }
-    return false;
-  });
+function addMarkerToMap(destLat, destLng, destName) {}
+
+/**
+ * Creates a map and adds it to the page.
+ * This function is implemented in another PR.
```
megankj, line 119:
```suggestion
 * TODO: Implement this function
```

src/main/webapp/go.js
```diff
@@ -103,18 +164,28 @@ function createLine(text) {
...
  * Show or hide the proceed button.
- *
  * @param {boolean} hide Whether the proceed button should be hidden or shown.
  */
 function toggleProceedButton(hide) {
   const proceedButton = document.getElementById(PROCEED_BUTTON);
   if (hide) {
     proceedButton.classList.add(INVISIBLE_CLASS);
   } else {
+    if (destIndex == huntArr.length - 1) {
```
megankj, line 174: You might want to use === instead of == here because it will enforce value and type equality
```suggestion
    if (destIndex === huntArr.length - 1) {
```

src/main/webapp/go.js
```diff
@@ -103,18 +164,28 @@ function createLine(text) {
...
+          PROCEED_FINAL_MSSG;
+    }
     proceedButton.classList.remove(INVISIBLE_CLASS);
   }
 }
 
+/**
+ * Show or hide the hint button.
+ * @param {boolean} hide Whether the proceed button should be hidden or shown.
+ * This function is implemented in the adjacent PR.
```
megankj, line 185:
```suggestion
 * TODO: Implement this function.
```

src/main/java/com/google/sps/servlets/GoDataServlet.java
```diff
@@ -42,7 +45,15 @@
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     String indexStr = request.getParameter(INDEX_PARAMETER);
-    index = Integer.parseInt(indexStr);
+    try {
+      index = Integer.parseInt(indexStr);
+    } catch (Exception e) {
+    }
+
+    nameGuess = request.getParameter(NAME_PARAMETER);
```
tjgoog, line 53: If I understand correctly, this design is that the client is expected to send only an index XOR a guess. Because those are two different pieces of functionality, they should be more clearly separated. I think separating them by having different URL paths for each would be the most simple and clear way of doing this*. I'm not sure how to configure a servlet to handle multiple different URL paths, but I imagine there's a way, otherwise you could write an additional servlet class for the different path. I can help you look for a way to do multiple paths in the same class later on if you'd like.

* Other ideas include having an additional parameter that specifies which set of functionality the client wants from the server for this request.

lilymzhou: This makes sense -- I think it'd be clearer if I split GoDataServlet into two separate servlets to handle getting the scavenger hunt data and getting the user's answer separately.

src/main/java/com/google/sps/data/ScavengerHunt.java
```diff
@@ -22,10 +22,12 @@
   // For saved scavenger hunts: represents which destination the user is currently looking for.
   private int index;
   private String city;
+  private String guess;
```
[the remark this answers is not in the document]
lilymzhou: I added this in primarily for the convenience of being able to easily convert all of the data sent to /go-data into one JSON object -- it's not really relevant to the ScavengerHunt data itself.

src/main/webapp/go.js
```diff
@@ -124,19 +174,43 @@ function hideStartButton() {
...
 /**
- * Change the riddle text displayed on the mmain page.
- * @param {String} text Text that the riddle should be changed to.
+ * Update {code@ display} to specified {code@ text}.
+ * @param {String} display ID of element to be updated
+ * @param {String} text Text that display should be updated to.
+ */
+function updateMessage(display, text) {
+  const message = document.getElementById(display);
+  if (display == HINT_DISPLAY) {
```
tjgoog, line 183:
```suggestion
  if (display == HINT_DISPLAY) {
  // Special case for HINT_DISPLAY to avoid clearing out its existing contents because it may already contain previous hints.
```

src/main/webapp/go.js
```diff
@@ -124,19 +174,43 @@ function hideStartButton() {
...
+}
+
+/**
+ * After the user has found all destinations, replace the riddle
+ * with a final message to the user.
  */
-function changeRiddleMessage(text) {
+function updateRiddleToFinalMessage() {
+  const newLine = document.createElement('div');
+  let message = '<p> Congrats! You\'ve visited the following locations:</p>';
```
tjgoog, line 198: Nice, I like that this shows all of the places (and sanitizes the content!)

src/main/webapp/go.js
```diff
@@ -44,49 +48,85 @@ window.onload = getHunt();
 function getHunt() {
   fetch(DATA_URL).then((response) => response.json()).then((mssg) => {
     destIndex = mssg.index;
-    for (let i = 0; i < mssg.items.length; i++) {
-      puzzleArr.push(mssg.items[i].riddle.puzzle);
-      destArr.push(mssg.items[i].name);
+    if (huntArr.length == 0) {
```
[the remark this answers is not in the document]
lilymzhou: huntArr should always be empty when getHunt() is called -- I'll remove the if loop here.

src/main/webapp/go.html
```diff
@@ -16,9 +16,9 @@
       <div class="center-text" id="map-message-area"></div>
       <form action="/guess-data" method="POST">
         <input type="text" name="guess-input" size="50"/>
-        <input type="submit" class="button" id="submit-button"/>
+        <input type="submit" class="submit-button" id="submit-button"/>
```
tjgoog, line 19: FYI I think you could replace the `input` tag with a `button`, so that it could automatically be styled by the CSS without having an extra class, like so:
```suggestion
        <button type="submit"/>
```
I'm fine with leaving this as-is though. Just sharing the info.

PR #23, opened 2020-08-03
Add functionality for map and hint button.

lilymzhou:
- Clicking the hint button displays a hint associated with the destination.
- Displays a map with the user's current location + all found destinations marked using markers

**The screenshots below incorporate changes from PR #22 .

At the start of the hunt:
P1

P2

After clicking the hint button twice:
P3

After correctly guessing the destination:
P4

At the end of the hunt:
P5

### DIFF
--- a/src/main/webapp/go.js
+++ b/src/main/webapp/go.js
@@ -33,7 +33,6 @@ const FINAL_MSSG = 'Congrats, you\'ve finished the hunt!';
 // Other constants.
 const INDEX_PARAM = 'new-index';
 const INVISIBLE_CLASS = 'invisible';
-const mapKey = config.MAP_KEY;
 
 // Global variables.
 const puzzleArr = [];
@@ -51,7 +50,7 @@ window.onload = function() {
  */
 function addScriptToHead() {
   const newScript = document.createElement('script');
-  newScript.src = 'https://maps.googleapis.com/maps/api/js?key=' + mapKey;
+  newScript.src = 'https://maps.googleapis.com/maps/api/js?key=' + config.MAP_KEY;
   document.getElementsByTagName('head')[0].appendChild(newScript);
 }
 
@@ -111,10 +110,12 @@ function createMap() {
   }
 }
 
-/** PR #2
+/**
  * Show or hide the hint button.
  * @param {boolean} hide Whether the proceed button should be hidden or shown.
- * Disable lint check because toggleHintButton() is called in PR #1.
+ * Disable lint check because this will later be called by 
+ * handleDestinationAnswer(), updateToCurrentState(), and proceed().
+ * 
  */
 function toggleHintButton(hide) { //eslint-disable-line
   const hintButton = document.getElementById(HINT_BUTTON);
--- a/src/main/webapp/go.js
+++ b/src/main/webapp/go.js
@@ -101,8 +101,9 @@ function createMap() {
       addMarkerToMap(position.coords.latitude, position.coords.longitude,
           'Your current location');
       map.setCenter(pos);
-    }, 
-    () => updateMessage(MAP_MSSG_DISPLAY, 'Error: I can\'t find your location.'));
+    },
+    () => updateMessage(MAP_MSSG_DISPLAY,
+        'Error: I can\'t find your location.'));
   } else {
     updateMessage(MAP_MSSG_DISPLAY, 'Error: Your browser doesn\'t' +
         'support geolocation.');
@@ -112,9 +113,9 @@ function createMap() {
 /**
  * Show or hide the hint button.
  * @param {boolean} hide Whether the proceed button should be hidden or shown.
- * Disable lint check because this will later be called by 
+ * Disable lint check because this will later be called by
  * handleDestinationAnswer(), updateToCurrentState(), and proceed().
- * 
+ *
  */
 function toggleHintButton(hide) { //eslint-disable-line
   const hintButton = document.getElementById(HINT_BUTTON);
--- a/src/main/webapp/go.js
+++ b/src/main/webapp/go.js
@@ -37,6 +37,7 @@ const INVISIBLE_CLASS = 'invisible';
 // Global variables.
 const puzzleArr = [];
 let destIndex;
+let hintIndex = 0; // Marks the hint that the user will see next.
 const destArr = [];
 let map;
 
@@ -65,7 +66,6 @@ function getHunt() {
       puzzleArr.push(mssg.items[i].riddle.puzzle);
       destArr.push(mssg.items[i].name);
     }
-    //createMap();
     updateToCurrentState(destIndex);
   });
 }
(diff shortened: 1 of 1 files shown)


src/main/webapp/go.js
```diff
@@ -21,21 +22,38 @@ const DATA_URL = '/go-data';
...
+  addScriptToHead();
+  getHunt();
+};
+
+/**
+ * Add the Map API key to the head.
+ */
+function addScriptToHead() {
+  const newScript = document.createElement('script');
+  newScript.src = 'https://maps.googleapis.com/maps/api/js?key=' + mapKey;
```
megankj, line 54: probably not much value in creating the mapKey variable... would just use config.MAP_KEY directly here.

src/main/webapp/go.js
```diff
@@ -48,10 +66,87 @@ function getHunt() {
...
+  if (navigator.geolocation) {
+    navigator.geolocation.getCurrentPosition(function(position) {
+      const pos = {lat: position.coords.latitude,
+        lng: position.coords.longitude};
+      addMarkerToMap(position.coords.latitude, position.coords.longitude,
+          'Your current location');
+      map.setCenter(pos);
+    }, function() {
+      updateMessage(MAP_MSSG_DISPLAY, 'Error: I can\'t find your location.');
+    });
```
megankj, line 107: This is more of a suggestion than a directive, but this might be a good opportunity to use [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) here!  Arrow functions are nice because they produce shorter code, and also handle the `this` binding differently.  If you're not familiar with javascript scope and the `this` keyword, you can read [this article](https://medium.com/tfogo/advantages-and-pitfalls-of-arrow-functions-a16f0835799e#:~:text=Arrow%20functions%20are%20best%20for,of%20this%20in%20their%20scope.).  Let me know if you run into any issues.

```suggestion
    navigator.geolocation.getCurrentPosition((position) => {
      const pos = {lat: position.coords.latitude,
        lng: position.coords.longitude};
      addMarkerToMap(position.coords.latitude, position.coords.longitude,
          'Your current location');
      map.setCenter(pos);
    }, 
    () => updateMessage(MAP_MSSG_DISPLAY, 'Error: I can\'t find your location.'));
```

src/main/webapp/go.js
```diff
@@ -48,10 +66,87 @@ function getHunt() {
...
+  } else {
+    updateMessage(MAP_MSSG_DISPLAY, 'Error: Your browser doesn\'t' +
+        'support geolocation.');
+  }
+}
+
+/** PR #2
```
megankj, line 114: What is this comment for?

src/main/webapp/go.js
```diff
@@ -48,10 +66,87 @@ function getHunt() {
...
+ * Show or hide the hint button.
+ * @param {boolean} hide Whether the proceed button should be hidden or shown.
+ * Disable lint check because toggleHintButton() is called in PR #1.
```
megankj, line 117: say where it's called in the code, not in what PR you introduced it

src/main/webapp/go.js
lilymzhou: oops -- I initially added comments to decide how to break up the pull requests, and forgot to take them out.

src/main/webapp/go.js
```diff
@@ -34,8 +38,21 @@ const INVISIBLE_CLASS = 'invisible';
...
+  addScriptToHead();
+  getHunt();
+};
+
+/**
+ * Add the Map API key to the head.
+ */
+function addScriptToHead() {
+  const newScript = document.createElement('script');
+  newScript.src = 'https://maps.googleapis.com/maps/api/js?key=' + config.MAP_KEY;
```
tjgoog, line 53: Please use the `callback` URL param in order to make sure the Maps API has loaded before using it. Perhaps the most straightforward is to set the callback to `createMap` and remove the existing `createMap()` call.

See documentation here: https://developers.google.com/maps/documentation/javascript/overview#Loading_the_Maps_API

@rysimone may be able to help with this if you run into any issues, since he used `callback` in the portfolio project.

src/main/webapp/go.js
```diff
@@ -48,10 +65,89 @@ function getHunt() {
...
+ * Creates a map and adds it to the page.
+ */
+function createMap() {
+  map = new google.maps.Map(
+      document.getElementById(MAP_DISPLAY),
+      // Centered at GooglePlex (updated below).
+      {center: {lat: 37.422, lng: -122.084}, zoom: 7},
+  );
+  if (navigator.geolocation) {
+    navigator.geolocation.getCurrentPosition((position) => {
```
tjgoog, line 98: I think it would be better to use `setMyLocationEnabled` so that the location is updated when the user moves. https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap#public-void-setmylocationenabled-boolean-enabled

src/main/webapp/go.js
```diff
@@ -48,10 +65,89 @@ function getHunt() {
...
+}
+
+/**
+ * Allow the user to see a hint.
+ * Disable lint check because getHint() is called from go.html.
+ */
+function getHint() { //eslint-disable-line
+  if (destIndex >= 0) {
+    const hintArr = huntArr[destIndex].hints;
+    if (hintIndex < hintArr.length) {
```
tjgoog, line 136: I don't see where hintIndex is declared. I _think_ that `hintIndex` is being implicitly initialized upon its first mention. Could you explicitly declare `hintIndex` near the top of the file, à la destIndex?

src/main/webapp/go.js
lilymzhou: I added a call to window.setInterval() that updates the user's location on the map every 10 seconds.

src/main/webapp/go.js
```diff
@@ -33,10 +33,12 @@ const FINAL_MSSG = 'Congrats, you\'ve finished the hunt!';
 // Other constants.
 const INDEX_PARAM = 'new-index';
 const INVISIBLE_CLASS = 'invisible';
+const REFRESH_TIME = 10000; //ten seconds
 
 // Global variables.
 const puzzleArr = [];
 let destIndex;
+let hintIndex = 0; // Marks the hint that the user will see next.
```
tjgoog, line 41: Should this be reset to 0 when the proceed to the next destination?

src/main/webapp/go.js
```diff
@@ -45,12 +47,14 @@ window.onload = function() {
   getHunt();
 };
 
+window.setInterval(updateGeolocation, REFRESH_TIME);
```
tjgoog, line 50: I think it would be better to do this inside the createMap method to make sure it won't fire before the maps API is ready

src/main/webapp/go.js
lilymzhou: The page reloads once the user submits an answer, so hintIndex is automatically set to 0 when the user begins to find the next destination.

PR #29, opened 2020-08-03
Creates an html page where users can input information to create a new destination for scavenger hunts

rysimone:


src/main/webapp/destinationCreation.html
```diff
@@ -0,0 +1,58 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8"/>
+    <title>Destination Creation</title>
+    <link rel="stylesheet" href="destinationCreation.css"/>
```
megankj, line 6: are you missing the css file for this PR?

rysimone: Oops. I forgot to add that file to this PR. I will add that in. Thanks

PR #30, opened 2020-08-03
Create a servlet for filtering, turn filters dark blue when clicked

shreyachatterjee00:
* Turn filters dark blue when clicked / light blue when unclicked
* Create a java servlet for filtering
* Pass an array of filters that the user has clicked on to the servlet (using doPost)

### DIFF
--- a/src/main/java/com/google/sps/servlets/GenerateServlet.java
+++ b/src/main/java/com/google/sps/servlets/GenerateServlet.java
@@ -33,7 +33,7 @@ public class GenerateServlet extends HttpServlet {
     // Get array of clicked filters and convert to ArrayList<String>
     Gson gson = new Gson();
     HashSet<String> clickedFilters = 
-      gson.fromJson(request.getParameter(FILTER_ARRAY), HashSet.class);
+        gson.fromJson(request.getParameter(FILTER_ARRAY), HashSet.class);
 
     response.setContentType("text/html;");
     response.getWriter().println(clickedFilters);
--- a/src/main/java/com/google/sps/servlets/GenerateServlet.java
+++ b/src/main/java/com/google/sps/servlets/GenerateServlet.java
@@ -32,7 +32,7 @@ public class GenerateServlet extends HttpServlet {
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     // Get array of clicked filters and convert to ArrayList<String>
     Gson gson = new Gson();
-    HashSet<String> clickedFilters = 
+    HashSet<String> clickedFilters =
         gson.fromJson(request.getParameter(FILTER_ARRAY), HashSet.class);
 
     response.setContentType("text/html;");
--- a/src/main/java/com/google/sps/servlets/GenerateServlet.java
+++ b/src/main/java/com/google/sps/servlets/GenerateServlet.java
@@ -32,7 +32,8 @@ public class GenerateServlet extends HttpServlet {
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     // Get array of clicked filters and convert to ArrayList<String>
     Gson gson = new Gson();
-    HashSet<String> clickedFilters = gson.fromJson(request.getParameter(FILTER_ARRAY), HashSet.class);
+    HashSet<String> clickedFilters = 
+      gson.fromJson(request.getParameter(FILTER_ARRAY), HashSet.class);
 
     response.setContentType("text/html;");
     response.getWriter().println(clickedFilters);
--- a/src/main/java/com/google/sps/servlets/GenerateServlet.java
+++ b/src/main/java/com/google/sps/servlets/GenerateServlet.java
@@ -23,22 +23,18 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 /** Servlet that returns bucket list content */
-@WebServlet("/generate")
-public class generateServlet extends HttpServlet {
+@WebServlet("/generate-hunt")
+public class GenerateServlet extends HttpServlet {
 
   private static final String FILTER_ARRAY = "clicked-array";
-  String JSONclickedArray;
-  HashSet<String> clickedFilters;
 
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     // Get array of clicked filters and convert to ArrayList<String>
-    JSONclickedArray = request.getParameter(FILTER_ARRAY);
     Gson gson = new Gson();
-    clickedFilters = gson.fromJson(JSONclickedArray, HashSet.class);
+    HashSet<String> clickedFilters = gson.fromJson(request.getParameter(FILTER_ARRAY), HashSet.class);
 
     response.setContentType("text/html;");
     response.getWriter().println(clickedFilters);
-    response.sendRedirect("/index.html");
   }
 }
--- a/src/main/java/com/google/sps/servlets/generateServlet.java
+++ b/src/main/java/com/google/sps/servlets/generateServlet.java
@@ -15,8 +15,8 @@
 package com.google.sps.servlets;
 
 import com.google.gson.Gson;
-import java.util.HashSet;
 import java.io.IOException;
+import java.util.HashSet;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
--- a/src/main/java/com/google/sps/servlets/generateServlet.java
+++ b/src/main/java/com/google/sps/servlets/generateServlet.java
@@ -1,0 +1,51 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.servlets;
+
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.util.ArrayList;
+import com.google.gson.Gson;
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Query.SortDirection;
+import java.util.HashSet;
+
+/** Servlet that returns bucket list content */
+@WebServlet("/generate")
+public class generateServlet extends HttpServlet {
+
+  private static final String FILTER_ARRAY = "clicked-array";
+  String JSONclickedArray;
+  HashSet<String> clickedFilters;
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Get array of clicked filters and convert to ArrayList<String>
+    JSONclickedArray = request.getParameter(FILTER_ARRAY);
+    Gson gson = new Gson();
+    clickedFilters = gson.fromJson(JSONclickedArray, HashSet.class);
+
+    response.setContentType("text/html;");
+    response.getWriter().println(clickedFilters);
+    response.sendRedirect("/index.html");
+  }
+}
--- a/src/main/java/com/google/sps/servlets/generateServlet.java
+++ b/src/main/java/com/google/sps/servlets/generateServlet.java
@@ -14,20 +14,13 @@
 
 package com.google.sps.servlets;
 
+import com.google.gson.Gson;
+import java.util.HashSet;
 import java.io.IOException;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-import java.util.ArrayList;
-import com.google.gson.Gson;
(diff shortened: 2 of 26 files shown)


megankj:
Removed @tjgoog and I for now - plz add back when the others approve.

src/main/webapp/filterScript.js
```diff
@@ -0,0 +1,39 @@
+/**
+* Turn a filter a different color when pressed, and change class accordingly.
+**/
+function turnBlueWhenClicked() { //eslint-disable-line
+  const allFilters = document.querySelectorAll('.unclicked-filter');
+  for (let i = 0; i < allFilters.length; i++) {
+    allFilters[i].onclick = function() {
+      const currFilter = allFilters[i];
+      if (currFilter.classList.contains('unclicked-filter')) {
```
lilymzhou, line 9: Since 'unclicked-filter' and 'clicked-filter' are used multiple times in this file, I suggest replacing these with constants.

src/main/webapp/filterScript.js
```diff
@@ -0,0 +1,39 @@
...
+function getClickedFilters() { //eslint-disable-line
+  const clickedArray = [];
+  const clickedFilters = document.querySelectorAll('.clicked-filter');
+  for (let i = 0; i < clickedFilters.length; i++) {
+    clickedArray[i] = clickedFilters[i].innerHTML;
+  }
+
+  const JSONArray = JSON.stringify(clickedArray);
+  console.log(JSONArray);
+  fetch('/generate?clicked-array=' + JSONArray, {method: 'POST'}).then(
```
lilymzhou, line 33: Similarly, I might make this url a const, just for clarity.

src/main/java/com/google/sps/servlets/generateServlet.java
```diff
@@ -0,0 +1,44 @@
...
+import java.io.IOException;
+import java.util.HashSet;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that returns bucket list content */
+@WebServlet("/generate")
+public class generateServlet extends HttpServlet {
```
megankj, line 27: Class names should be UpperCamelCase.  Please rename the class and the file.

src/main/java/com/google/sps/servlets/generateServlet.java
```diff
@@ -0,0 +1,44 @@
...
+  private static final String FILTER_ARRAY = "clicked-array";
+  String JSONclickedArray;
+  HashSet<String> clickedFilters;
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Get array of clicked filters and convert to ArrayList<String>
+    JSONclickedArray = request.getParameter(FILTER_ARRAY);
+    Gson gson = new Gson();
+    clickedFilters = gson.fromJson(JSONclickedArray, HashSet.class);
```
megankj, line 38: The doPost method is the only place this variable is used, so no need to make this a class variable. In fact, you don't need a local variable at all.  Same for clickedFilters.  If it's not used elsewhere, you should contain the scope of variables to the methods they are used in.
 
```suggestion
    HashSet<String> clickedFilters = gson.fromJson(request.getParameter(FILTER_ARRAY), HashSet.class);
```

src/main/java/com/google/sps/servlets/generateServlet.java
```diff
@@ -0,0 +1,44 @@
...
+public class generateServlet extends HttpServlet {
+
+  private static final String FILTER_ARRAY = "clicked-array";
+  String JSONclickedArray;
+  HashSet<String> clickedFilters;
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Get array of clicked filters and convert to ArrayList<String>
+    JSONclickedArray = request.getParameter(FILTER_ARRAY);
```
megankj, line 36: non constant variables shouldn't start with a capital letter like this

src/main/webapp/filterScript.js
```diff
@@ -0,0 +1,43 @@
+const UNCLICKED = 'unclicked-filter';
+const CLICKED = 'clicked-filter';
+const CLICKED_ARRAY_URL = 'clicked-array=';
+
+/**
+* Turn a filter a different color when pressed, and change class accordingly.
+**/
+function turnBlueWhenClicked() { //eslint-disable-line
+  const allFilters = document.querySelectorAll('.unclicked-filter');
```
megankj, line 9:
```suggestion
  const allFilters = document.querySelectorAll('.' + UNCLICKED);
```

src/main/webapp/filterScript.js
```diff
@@ -0,0 +1,43 @@
...
+    allFilters[i].onclick = function() {
+      const currFilter = allFilters[i];
+      if (currFilter.classList.contains(UNCLICKED)) {
+        currFilter.classList.remove(UNCLICKED);
+        currFilter.classList.add(CLICKED);
+      } else {
+        currFilter.classList.remove(CLICKED);
+        currFilter.classList.add(UNCLICKED);
+      }
+    };
```
megankj, line 20: No need to change it, but if you haven't seen [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) in javascript before, let me introduce you to them! Arrow functions are nice because they produce shorter code, and also handle the `this` binding differently. If you're not familiar with javascript scope and the `this` keyword, you can read [this article](https://medium.com/tfogo/advantages-and-pitfalls-of-arrow-functions-a16f0835799e#:~:text=Arrow%20functions%20are%20best%20for,of%20this%20in%20their%20scope.).

src/main/webapp/filterScript.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+}
+
+/**
+* Get all filters that have been clicked when user presses submit button,
+* and pass array to servlet, and TODO: return success or error message.
+**/
+function getClickedFilters() { //eslint-disable-line
+  const clickedArray = [];
+  const clickedFilters = document.querySelectorAll('.clicked-filter');
```
megankj, line 30:
```suggestion
  const clickedFilters = document.querySelectorAll('.' + CLICKED);
```

src/main/webapp/filterScript.js
```diff
@@ -0,0 +1,43 @@
...
+  for (let i = 0; i < clickedFilters.length; i++) {
+    clickedArray[i] = clickedFilters[i].innerHTML;
```
megankj, line 32: innerText might be better?  What if you add something later that makes innerHTML not just text?

src/main/webapp/filterScript.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  const JSONArray = JSON.stringify(clickedArray);
```
megankj, line 35: non-constant variables shouldn't start with a capital letter

src/main/java/com/google/sps/servlets/generateServlet.java
```diff
@@ -0,0 +1,44 @@
...
+    Gson gson = new Gson();
+    clickedFilters = gson.fromJson(JSONclickedArray, HashSet.class);
+
+    response.setContentType("text/html;");
+    response.getWriter().println(clickedFilters);
+    response.sendRedirect("/index.html");
```
tjgoog, line 42: Sending a redirect will effectively overwrite the response that you wrote. I think you should just remove this line.

Edit: Or maybe the response text will still come through and be used when this is accessed via `fetch()`, based on code down below. Regardless, writing content to the response and sending a redirect shouldn't be done with the same response, so I still think this line should be removed.

src/main/webapp/filterScript.js
```diff
@@ -0,0 +1,43 @@
+const UNCLICKED = 'unclicked-filter';
```
tjgoog, line 1: nit: no need to include `Script` in the file name since the `.js` extension already says that it's a script.

src/main/webapp/filterScript.js
```diff
@@ -0,0 +1,43 @@
...
+const CLICKED = 'clicked-filter';
+const CLICKED_ARRAY_URL = 'clicked-array=';
+
+/**
+* Turn a filter a different color when pressed, and change class accordingly.
```
tjgoog, line 6: Because the class change is what makes the filter color change, I think that should be the emphasis. Something like "swap between setting the clicked and unclicked classes (which will change the color of the button) on a button when it is clicked."

src/main/webapp/filterScript.js
```diff
@@ -0,0 +1,43 @@
...
+**/
+function turnBlueWhenClicked() { //eslint-disable-line
```
tjgoog, line 8: Add a note for why lint is disabled- e.g. called from an `onclick` attribute in HTML, or to-be used in a future PR.

src/main/webapp/filterScript.js
```diff
@@ -0,0 +1,43 @@
...
+        currFilter.classList.remove(CLICKED);
+        currFilter.classList.add(UNCLICKED);
+      }
+    };
+  }
+}
+
+/**
+* Get all filters that have been clicked when user presses submit button,
+* and pass array to servlet, and TODO: return success or error message.
```
tjgoog, line 26: I think `sendClickedFiltersToServer` or similar would be a better name, since `getClickedFilters` suggests that this method will return the collection of filters that have been clicked.

src/main/webapp/filterScript.js
```diff
@@ -0,0 +1,43 @@
+const UNCLICKED = 'unclicked-filter';
+const CLICKED = 'clicked-filter';
+const CLICKED_ARRAY_URL = 'clicked-array=';
```
tjgoog, line 3: The name of this constant should include the information that this is the name of a parameter key. So `CLICKED_ARRAY_URL_PARAM` or `CLICKED_ARRAY_KEY` or something would be more clear. Also, it's more common to keep the `=` out of the constant value in situations like this and just apply it when used, like `CLICKED_ARRAY_KEY + '=' value`. This design gives you more flexibility, e.g. you could use the same constant when reading data out of a URL (not relevant here, but does come up in some situations) or using a more structured URL creation method.

src/main/webapp/filterScript.js
```diff
@@ -0,0 +1,43 @@
...
+function getClickedFilters() { //eslint-disable-line
+  const clickedArray = [];
+  const clickedFilters = document.querySelectorAll('.clicked-filter');
+  for (let i = 0; i < clickedFilters.length; i++) {
+    clickedArray[i] = clickedFilters[i].innerHTML;
+  }
+
+  const JSONArray = JSON.stringify(clickedArray);
+  fetch('/generate?' + CLICKED_ARRAY_URL + JSONArray, {method: 'POST'}).then(
+      (response) => response.json()).then((message) => {
```
tjgoog, line 37: I think this should be reading text rather than json since the response type is plain text.
```suggestion
      (response) => response.text()).then((message) => {
```

src/main/webapp/filterStyle.css
```diff
@@ -65,8 +65,23 @@
   width: 300px;
 }
 
-.filter {
-  background: #69d1dd;
+.unclicked-filter {
```
tjgoog, line 68: I think you should keep the `.filter` rule and class and keep the set of attributes for `.unclicked-filter` and `.clicked-filter` minimal, just overriding the background color (could even make the unclicked background part of `.filter` and just override it for `.clicked-filter`*)

* This may get in to [CSS specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity).

src/main/webapp/filterScript.js
shreyachatterjee00: I have seen arrow functions! I think I'm just more comfortable with this syntax since it's more like declaring a function how I would normally, but I can change it to arrow syntax if it's better in this case

src/main/webapp/filter.js
```diff
@@ -0,0 +1,50 @@
+const FILTER = '.filter';
+const UNCLICKED = 'unclicked-filter';
+const CLICKED = 'clicked-filter';
+const CLICKED_ARRAY_URL_PARAM = 'clicked-array';
+
+/**
+* Swap between "clicked" and "unclicked" classes
+* (which change the color of the button) when the button is pressed.
+* Note: function declaration disabled because the function is called
```
tjgoog, line 9:
```suggestion
* Note: function declaration lint disabled because the function is called
```

src/main/webapp/filter.js
```diff
@@ -0,0 +1,50 @@
...
+**/
+function sendClickedFiltersToServer() { //eslint-disable-line
+  const clickedArray = [];
+  const clickedFilters = document.querySelectorAll('.' + CLICKED);
+  for (let i = 0; i < clickedFilters.length; i++) {
+    clickedArray[i] = clickedFilters[i].innerText;
+  }
+
+  const jsonArray = JSON.stringify(clickedArray);
+  fetch('/generate-hunt?' + CLICKED_ARRAY_URL_PARAM + '=' + jsonArray,
```
tjgoog, line 42: 👍 This is a good example of the sort of request design we were talking about today.

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -0,0 +1,41 @@
...
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Get array of clicked filters and convert to ArrayList<String>
+    Gson gson = new Gson();
+    HashSet<String> clickedFilters =
+        gson.fromJson(request.getParameter(FILTER_ARRAY), HashSet.class);
+
+    response.setContentType("text/html;");
+    response.getWriter().println(clickedFilters);
```
megankj, line 39: What is the longer-term plan for this return value?  Right now you are just returning what was requested, but eventually you'll want to start the scavenger hunt if you were able to create one.  Is the plan that you get the filters, query datastore for destinations, pick some, and create a scavenger hunt, store it in datastore and... return an ID? return the scavenger hunt?  It's my understanding that @lilymzhou is working on querying the datastore for a hunt, but if her code also has the client requesting the code from the datastore, wouldn't it make more sense to just return the hunt directly from this request instead (what I suspect might be happening) only using this POST request to store the data and another GET request to get the hunt you just created from the datastore?

I want to make sure these two pieces are actually connected together and not just tangentially related via datastore.  Thanks!

megankj: Chatted offline - @shreyachatterjee00 and @lilymzhou should meet to discuss how to make this one request flow instead of 2.

PR #33, opened 2020-08-04
Servlet that retrieves user submitted information and creates a destination from this data

rysimone:


src/main/java/com/google/sps/data/Destination.java
```diff
@@ -0,0 +1,113 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+ 
+package com.google.sps.data;
+ 
+import java.lang.Enum;
+import java.util.ArrayList;
+import java.util.List;
```
lilymzhou, line 19: It seems like List is unused in this class.

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.stream.Collectors;
+import java.util.Set;
+import java.util.HashSet;
+
+/* Returns a destination created from user submitted information */
+@WebServlet("/destination-data")
+public class DestinationDataServlet extends HttpServlet {
+  
+  // Temporarily stores the destination created by the user
+  public List<com.google.sps.data.Destination> destinations = new ArrayList<>();
```
lilymzhou, line 35: I believe you can actually write `import com.google.sps.data.Destination;` at the top, and use `List<Destination> destinations` throughout the rest of this file (and same for Location, Riddle, etc.) -- this will make the code somewhat cleaner.

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -0,0 +1,133 @@
...
+            .withLocation(location)
+            .withCity(city)
+            .withDescription(description)
+            .withRiddle(riddle)
+            .withTags(checkedTags)
+            .withObscurity(level)
+            .build();
+    
+    destinations.add(d1);
+    response.sendRedirect("/destination-data");
```
lilymzhou, line 75: Should this be redirecting to the .html page instead? From what I remember, sending a POST request to /destination-data will redirect to this URL by default.

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
rysimone: Thanks I was wondering if there was a way for me to do that.

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
rysimone: I redirect to /destination-data just to show that the information has been retrieved. I plan to change it to redirect to a .html once datastore is implemented.

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
shreyachatterjee00: ^^ Agree with Lily's comment here!

src/main/java/com/google/sps/data/Destination.java
```diff
@@ -0,0 +1,114 @@
...
+    MEDIUM,
+    HARD,
+    UNDEFINED;
+  }
+
+  private String name;
+  private LatLng location;
+  private String city;
+  private String description;
+  private ArrayList<com.google.sps.data.Riddle> riddles = new ArrayList<>();
```
megankj, line 45: If two classes are in the same package (in this case, `com.google.sps.data`), you don't need to prefix the class name like this, nor do you need to add an import statement at the top.  DestinationDataServlet, by contrast, is in the `com.google.sps.servlets` package, so you _do_ need to add the import statement in that case.

```suggestion
  private ArrayList<Riddle> riddles = new ArrayList<>();
```

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.stream.Collectors;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/* Returns a destination created from user submitted information */
+@WebServlet("/destination-data")
+public class DestinationDataServlet extends HttpServlet {
+
```
megankj, line 36:
```suggestion
  private static final NAME_PARAMETER = "name";
```

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -0,0 +1,147 @@
...
+  // Temporarily stores the destination created by the user
+  public List<Destination> destinations = new ArrayList<>();
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String name = request.getParameter("name");
```
megankj, line 42: Create constants for these parameter names and declare them at the top, like I did in the suggestion above.  Do this for all parameter names.

```suggestion
    String name = request.getParameter(NAME_PARAMETER);
```

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -0,0 +1,147 @@
...
+        new Riddle.Builder()
+            .withPuzzle(request.getParameter("riddle"))
+            .withHint(request.getParameter("hint1"))
+            .withHint(request.getParameter("hint2"))
+            .withHint(request.getParameter("hint3"))
+            .build();
+
+    /* Retrieves the obscurity level chosen by the user as a List of Strings and converts the level to an Enum Obscurity value */
+    List<String> levels =
+        Arrays.stream(request.getParameterValues("obscurity"))
```
megankj, line 63: this implies there can be more than 1 level of obsurity?

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -0,0 +1,147 @@
...
+    Destination.Obscurity level = convertLevelToEnum(levels);
+
+    /* Retrieves the tags selected by the user as a List of Strings and converts them into a Set of Enum Tags */
+    List<String> tags =
+        Arrays.stream(request.getParameterValues("tag"))
+            .filter(tag -> tag != null)
+            .collect(Collectors.toList());
+    Set<Destination.Tag> checkedTags = convertTagsToEnum(tags);
+
+    Destination d1 =
```
megankj, line 75: the doPost method will only ever create 1 destination at a time, so you can just call this variable `destination`

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -0,0 +1,147 @@
...
+            .withName(name)
+            .withLocation(location)
+            .withCity(city)
+            .withDescription(description)
+            .withRiddle(riddle)
+            .withTags(checkedTags)
+            .withObscurity(level)
+            .build();
+
+    destinations.add(d1);
```
megankj, line 86: The idea is that eventually this would write to the datastore, right?  If so, please leave a TODO comment so that it's clear what the ulitmate goal of this code is.

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
megankj: leave a TODO comment to indicate that you're planning on changing it to redirect to an HTML file

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -0,0 +1,147 @@
...
+    response.sendRedirect("/destination-data");
+  }
+
+  /* Retrieves the destination created by the user, turns it into a JSON formatted string,
```
megankj, line 90: This actually just retrieves the last destination that was created, by any person.  Update the comment to reflect what this method is doing, or leave a TODO to indicate what you plan to do.

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -0,0 +1,147 @@
...
+          tagEnums.add(Destination.Tag.FOOD);
+          break;
+        case "family":
+          tagEnums.add(Destination.Tag.FAMILY);
+          break;
+        case "tourist":
+          tagEnums.add(Destination.Tag.TOURIST);
+          break;
+        default:
+          tagEnums.add(Destination.Tag.UNDEFINED);
```
megankj, line 124: i don't think you want to add UNDEFINED to your set of tags, right?

rysimone:
### TODO
Store the user input information in datastore rather than locally.
Redirect to a .html file rather than the /destination-data.

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
rysimone: yes that is the plan

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
rysimone: That is correct. From what I remember from our conversation last Friday, you had said that this was a good practice for Enums. I can remove this if that is not the case.

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
rysimone: I see the confusion. I will give it a more fitting name;

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
megankj: Leave a comment in the code, not the PR. :)

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
megankj: Yes, it's good practice to have an UNDEFINED value when you define the enum.

In this particular case, you're taking in a list of strings the user has given you and converting it to a set of enums.  Since the user isn't ever going to be specifying "undefined" as a tag, there's no point in adding it to your enum set.

rysimone: Oh ok that makes sense. I will remove it

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
megankj: I imagine you'd just want to use getParameterValue instead of getParameterValues, right?

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -0,0 +1,147 @@
...
+      }
+    }
+
+    return tagEnums;
+  }
+
+  public Destination.Obscurity convertLevelToEnum(List<String> levels) {
```
megankj: Also I missed this, but this should just be a string with the 1 level, right?  Not a list of strings.

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
rysimone: Ok hahahaha

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
rysimone: Yes actually that works better for this use case.

rysimone: there is no getParameterValues method for the HttpServletRequest type so I am only storing the first value in the array since there will only be one

megankj: My mistake, the method is getParameter:
https://docs.oracle.com/javaee/6/api/javax/servlet/ServletRequest.html#getParameter(java.lang.String)

rysimone: Ok I got that one to work.

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -0,0 +1,152 @@
...
+   */
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    Gson gson = new Gson();
+    String json = gson.toJson(destinations.get(destinations.size() - 1));
+    response.setContentType("application/json;");
+    response.getWriter().println(json);
+  }
+
+  public Set<Destination.Tag> convertTagsToEnum(List<String> tags) {
```
megankj, line 112: this method should be private, not public, since it's not used by any other classes

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -0,0 +1,152 @@
...
+        case "tourist":
+          tagEnums.add(Destination.Tag.TOURIST);
+          break;
+      }
+    }
+
+    return tagEnums;
+  }
+
+  public Destination.Obscurity convertLevelToEnum(String obscureLevel) {
```
megankj, line 140: same with this one - make it private

src/main/java/com/google/sps/data/Destination.java
```diff
@@ -0,0 +1,114 @@
...
+
+    public Destination build() {
+      Destination destination = new Destination();
+      destination.name = this.name;
+      destination.location = this.location;
+      destination.city = this.city;
+      destination.description = this.description;
+      for (Riddle riddle : this.riddles) {
+        destination.riddles.add(riddle);
+      }
```
megankj, line 105: There's a shortcut for this, the addAll method:
https://docs.oracle.com/javase/7/docs/api/java/util/ArrayList.html#addAll(java.util.Collection)

```suggestion
      destination.riddles.addAll(this.riddles);
```

src/main/java/com/google/sps/data/Destination.java
```diff
@@ -0,0 +1,114 @@
...
+      for (Tag tag : this.categories) {
+        destination.categories.add(tag);
+      }
```
megankj, line 108: same here!
```suggestion
      destination.categories.addAll(this.categories);
```

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -0,0 +1,152 @@
...
+    switch (obscureLevel) {
+      case "easy":
+        return Destination.Obscurity.EASY;
+      case "medium":
+        return Destination.Obscurity.MEDIUM;
+      case "hard":
+        return Destination.Obscurity.HARD;
+    }
+
+    return Destination.Obscurity.UNDEFINED;
```
megankj, line 150: You can add a default case which will cover everything else
```suggestion
    switch (obscureLevel) {
      case "easy":
        return Destination.Obscurity.EASY;
      case "medium":
        return Destination.Obscurity.MEDIUM;
      case "hard":
        return Destination.Obscurity.HARD;
      default:
        return Destination.Obscurity.UNDEFINED; 
    }
```

megankj:
Please address the changes I left, but don't wait for Tim's review to merge this in.  Go ahead as soon as you've made the changes I've suggested.

PR #37, opened 2020-08-07
Change cloud project ID to be .googleplex.com project

tjgoog:
Change cloud project ID in pom.xml to be the new one that's used for the .googleplex.com version


tjgoog:
Not ready to submit yet because I need to get an App Engine app set up under this new project ID first. Removed reviewer for now.

PR #38, opened 2020-08-07
Basic filter

shreyachatterjee00:
* Filter by Place and Difficulty level
* Outputs (prints) a list of Destination objects that match filters (from Destinations I created)
* Add link to Filtering on index.html

Filtering: 
image

Output on console: 
image

Filtering: 
image

Output on console: 
image

(Current created destination objects - hard coded data): 
Golden Gate (Easy), Japanese Tea Garden (Medium), Orpheum Theater (Hard), Louvre (Easy), Eiffel Tower (Medium), Arc de Triomphe (Hard)


src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -28,14 +30,171 @@
...
+  }
+
+  /* Return Destination objects within specified place. */
+  public ArrayList<String> filterPlaces(
+      HashSet<String> userPlaces, ArrayList<Destination> allDestinations) {
+    ArrayList<String> filteredDestinations = new ArrayList();
+    // Iterate through destinations
+    // If Destination place is in userPlaces array
+    // Add Destination to filteredDestinations array
+    for (int i = 0; i < allDestinations.size(); i++) {
```
rysimone, line 90:
```suggestion
    for (Destination destination : allDestinations) {
```

rysimone: This for loop will iterate through the List while also assigning each value in the List to the new Destination variable you create in the for loop.

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -28,14 +30,171 @@
...
+      Destination currDestination = allDestinations.get(i);
+      if (userPlaces.contains(currDestination.getCity())) {
```
rysimone, line 92:
```suggestion
      if (userPlaces.contains(destination.getCity())) {
```

rysimone: so you can do this instead of creating a new variable to assign the current destination to

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -28,14 +30,171 @@
...
+  /* Return HashSet of only difficulty filters. */
+  public HashSet<String> separateDifficultyFilters(HashSet<String> clickedFilters) {
+    HashSet<String> userDifficulties = new HashSet();
+    for (int i = 0; i < allDifficulties.length; i++) {
+      if (clickedFilters.contains(allDifficulties[i])) {
+        userDifficulties.add(allDifficulties[i]);
+      }
+    }
+    return userDifficulties;
+  }
```
lilymzhou, line 81: Since separatePlaceFilters() and separateDifficultyFilters() are almost identical in structure, I suggest replacing these with a generic function (i.e. `separateFilters(HashSet<String> clickedFilters, HashSet<String> set)`).

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -28,14 +30,162 @@
...
+    filteredDestinations = filterPlaces(userPlaces, allDestinations);
+    filteredDestinations =
+        filterDifficulty(filteredDestinations, userDifficulties, allDestinations);
+    writeToDataStore(filteredDestinations);
 
     response.setContentType("text/html;");
     response.getWriter().println(clickedFilters);
   }
+
+  public HashSet<String> separateFilters(HashSet<String> clickedFilters, String[] allTypeFilters) {
```
tjgoog, line 61: Is this essentially taking the two inputs and returning the values that appear in both? I think that behavior can be implemented more simply by using [Set.retainAll()](https://docs.oracle.com/javase/7/docs/api/java/util/Set.html#retainAll(java.util.Collection)). Because that method takes a `Collection` rather than a primitive array, you can convert the input array to a `List` (which is a sub-interface of `Collection`) with [Arrays.asList()](https://docs.oracle.com/javase/7/docs/api/java/util/Arrays.html#asList(T...)), either in this method or by calling asList() when initializing the `allPlaces` and `allDifficulties` values at the top.

If you use Set.retainAll(), you may decide to remove this method and do that call directly from the places where this method is currently used, or you may decide it's more readable to keep this separate method. If you do keep this separate method, could you modify the name and/or add a Javadoc to more specifically explain what it does? e.g. `sanitizeFilters` or `restrictFiltersToKnownValues`.

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -28,14 +30,162 @@
...
+    HashSet<String> userFilters = new HashSet();
+    for (int i = 0; i < allTypeFilters.length; i++) {
+      if (clickedFilters.contains(allTypeFilters[i])) {
+        userFilters.add(allTypeFilters[i]);
+      }
+    }
+    return userFilters;
+  }
+
+  /* Return Destination objects within specified place. */
```
tjgoog, line 71: Nit: Is returning the names destinations rather than a full destination object, update javadoc.

src/main/java/com/google/sps/servlets/GenerateServlet.java
megankj: +1 to Tim's comment, I think you can remove this method altogether

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -28,14 +30,162 @@
...
 
+  String[] allPlaces = {"Paris", "New York City", "San Francisco", "London", "Sydney", "Venice"};
+  String[] allDifficulties = {"Easy", "Medium", "Hard"};
+
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    HashSet<String> userPlaces = new HashSet();
+    HashSet<String> userDifficulties = new HashSet();
+
+    ArrayList<Destination> allDestinations = new ArrayList();
```
megankj, line 41:
```suggestion
    ArrayList<Destination> allDestinations = createFakeDestinations();
```

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -28,14 +30,162 @@
...
     // Get array of clicked filters and convert to ArrayList<String>
-    Gson gson = new Gson();
     HashSet<String> clickedFilters =
-        gson.fromJson(request.getParameter(FILTER_ARRAY), HashSet.class);
+        new Gson().fromJson(request.getParameter(FILTER_ARRAY), HashSet.class);
+
+    allDestinations = createFakeDestinations();
+    userPlaces = separateFilters(clickedFilters, allPlaces);
+    userDifficulties = separateFilters(clickedFilters, allDifficulties);
+    filteredDestinations = filterPlaces(userPlaces, allDestinations);
```
megankj, line 52: For readability, I would suggest putting the variable declarations in-line with where they are assigned.  There isn't any value in declaring your variables as empty arrays and then assigning them values 9 lines down.   It makes it hard to follow the logic. 

```suggestion
    ArrayList<String> filteredDestinations = filterPlaces(userPlaces, allDestinations);
```

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -28,14 +30,162 @@
...
+    ArrayList<String> filteredDestinations = new ArrayList();
+
     // Get array of clicked filters and convert to ArrayList<String>
-    Gson gson = new Gson();
     HashSet<String> clickedFilters =
-        gson.fromJson(request.getParameter(FILTER_ARRAY), HashSet.class);
+        new Gson().fromJson(request.getParameter(FILTER_ARRAY), HashSet.class);
+
+    allDestinations = createFakeDestinations();
+    userPlaces = separateFilters(clickedFilters, allPlaces);
```
megankj, line 50:
```suggestion
    HashSet<String> userPlaces = separateFilters(clickedFilters, allPlaces);
```

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -28,14 +30,162 @@
...
+    userDifficulties = separateFilters(clickedFilters, allDifficulties);
```
megankj, line 51:
```suggestion
    HashSet<String> userDifficulties = separateFilters(clickedFilters, allDifficulties);
```

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -28,14 +30,162 @@
...
+      if (clickedFilters.contains(allTypeFilters[i])) {
+        userFilters.add(allTypeFilters[i]);
+      }
+    }
+    return userFilters;
+  }
+
+  /* Return Destination objects within specified place. */
+  public ArrayList<String> filterPlaces(
+      HashSet<String> userPlaces, ArrayList<Destination> allDestinations) {
```
megankj, line 73: I think you can actually just do this with streams and get rid of the method!

allDestinations.stream().filter(destination -> userPlaces.contains(destination.getCity()).map(destination -> destination.getName()).collect(toSet())

This filters the destinations to the ones in userPlaces and then maps the Destination object to it's name, collecting everything into a set

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -28,14 +30,162 @@
...
+      }
+    }
+    return filteredDestinations;
+  }
+
+  /* Return Destination objects with specified difficulty. */
+  public ArrayList<String> filterDifficulty(
+      ArrayList<String> filteredDestinations,
+      HashSet<String> userDifficulties,
+      ArrayList<Destination> allDestinations) {
```
megankj, line 90: I'm a little confused by the complexity of this approach, because I think it would make sense to simply use streams again here.

filteredDestinations.stream().filter(destination -> userDifficulties .contains(destination. getDifficulty());

this requires some small changes to your existing code:
- filteredDestinations should be a list of _Destination_s and not names
-  userDifficulties should be a set of Destination.Obscurity levels that correspond to what the user is okay with.  If they are okay with everything it should be full of every level, not empty

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -15,8 +15,12 @@
 package com.google.sps.servlets;
 
 import com.google.gson.Gson;
+import com.google.sps.data.Destination;
 import java.io.IOException;
+import java.util.*;
```
megankj, line 20: I think this might be importing everything in java.util.  Is there a reason you added this?

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -28,14 +32,143 @@
...
+  String[] allPlaces = {"Paris", "New York City", "San Francisco", "London", "Sydney", "Venice"};
+  String[] allDifficulties = {"Easy", "Medium", "Hard"};
+
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     // Get array of clicked filters and convert to ArrayList<String>
-    Gson gson = new Gson();
     HashSet<String> clickedFilters =
-        gson.fromJson(request.getParameter(FILTER_ARRAY), HashSet.class);
+        new Gson().fromJson(request.getParameter(FILTER_ARRAY), HashSet.class);
```
megankj, line 42: I think using the same request parameter to capture all filter arguments is going to give you some trouble.  From the looks of this code, it seems like the URL request might look like this:
/generate-hunt?clicked-array=[Paris,easy,medium,4,art,kid_friendly]

In this method, you are having to take all your filters, and see which ones pertain to Place vs. Difficulty by using retainAll to see what matches.  That might be a problem if you have something like a number since you won't know what it's for if you have multiple parameters that could be numbers.

But what if the request looked like this:
/generate-hunt?city=Paris&difficulty=easy,medium&num_stops=4&labels=art,kid_friendly

Then you don't need to do retainAll because you already know which ones the user chose.

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -28,14 +32,143 @@
...
+            userDifficultyLevels.add(Destination.Obscurity.MEDIUM);
+            break;
+          case "Hard":
+            userDifficultyLevels.add(Destination.Obscurity.HARD);
+            break;
+        }
+      }
+    }
+    return userDifficultyLevels;
+  }
```
megankj, line 93: I'm seeing some overlap between this method and the ones that @rysimone created in DestinationDataServlet
https://github.com/googleinterns/step183-2020/blob/master/src/main/java/com/google/sps/servlets/DestinationDataServlet.java#L140

I think it would probably be a good idea to create either a Utils class that could share some of this converter logic or create a static method in the Destination class that could covert a String to a Destination.Obscurity / Destination.Tag enum.

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -28,14 +32,143 @@
...
+        allDestinations.stream()
+            .filter(destination -> userPlaces.contains(destination.getCity()))
+            .map(destination -> destination)
+            .collect(Collectors.toSet());
+    // Filter by difficulty
+    filteredDestinations =
+        filteredDestinations.stream()
+            .filter(destination -> userDifficultyLevels.contains(destination.getDifficulty()))
+            .map(destination -> destination)
+            .collect(Collectors.toSet());
```
megankj, line 110: Two things:

1) This can be combined!  You want to filter by city _and_ difficulty here, so you can just specify that in the filter function.

2) The map method will transform the left hand side into the right hand side, i.e. if you did map(destination -> destination.getName()) it would turn your set of Destinations into a set of Strings.  Since you're not transforming destination, you don't need the map() part.

```suggestion
    Set<Destination> filteredDestinations =
        allDestinations.stream()
            .filter(destination -> userPlaces.contains(destination.getCity()) && userDifficultyLevels.contains(destination.getDifficulty()))
            .collect(Collectors.toSet());
```

src/main/java/com/google/sps/servlets/GenerateServlet.java
shreyachatterjee00: I created a static method in the Destination class to do this!

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -26,16 +30,116 @@
...
+        new Gson().fromJson(request.getParameter(DIFF_FILTERS), HashSet.class);
+
+    // Create fake destinations TODO: get destinations from datastore
+    ArrayList<Destination> allDestinations = createFakeDestinations();
+
+    // Convert difficulty level strings to Destination.Obscurity
+    HashSet<Destination.Obscurity> userDifficultyLevels = new HashSet();
+    for (String level : userDifficultyStrings) {
+      userDifficultyLevels.add(Destination.stringToEnum(level));
+    }
```
megankj, line 54: You probably don't want to add UNDEFINED to your set, so you should probably check it before adding it to userDifficultyLevels.

Also you don't have to change it since what's below might be more difficult to read, but this can also be done with streams (my fav!): 
`HashSet <Destination.Obscurity> userDifficultyLevels = userDifficultyStrings.stream().map(Destination:: stringToEnum).filter(obscurity -> obscurity != Destination.Obscurity.UNDEFINED).collect(Collectors.toSet());`

What this is doing:
1.   Creating a stream so you can process each item
2.  Map the strings to the corresponding Obscurity level
3.  filtering out the UNDEFINED enums
4.  collecting it back into a set

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -26,16 +30,116 @@
...
+    if (userDifficultyStrings.size() == 0) {
+      userDifficultyLevels.add(Destination.stringToEnum("Easy"));
```
megankj, line 56: I think this should go before you process userDifficultyStrings.

if (userDifficultyStrings.[isEmpty](https://docs.oracle.com/javase/7/docs/api/java/util/HashSet.html#isEmpty())()) {
    userDifficultyLevels.add(Destination.Obsurity.EASY);
    //etc.
} else {
    // process userDifficultyStrings
}

I think it's clearer to handle the "no preference" case first, since it's a special case of userDifficultyStrings.  Additionally, there's no need to use stringToEnum because you can just use the enum directly.

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -26,16 +30,116 @@
...
+
+    writeToDataStore(filteredDestinations);
 
     response.setContentType("text/html;");
-    response.getWriter().println(clickedFilters);
+    response.getWriter().println(filteredDestinations);
+  }
+
+  /* Return ArrayList<Destination> of filtered Destination objects. */
+  public Set<Destination> filter(
```
megankj, line 72: Now that we're adding more complicated filtering logic here, I think you should write tests to make sure your filter function is working as intended!  Can you add a test file with a couple tests to demonstrate that it returns the expected Destinations?

src/test/java/com/google/sps/GenerateHuntTest.java
```diff
@@ -0,0 +1,149 @@
...
+    HashSet<String> userPlaces = new HashSet();
+    userPlaces.add("San Francisco");
+    HashSet<Destination.Obscurity> userDiff = new HashSet();
+    userDiff.add(Destination.Obscurity.MEDIUM);
+
+    GenerateServlet generate = new GenerateServlet();
+    Set<Destination> actual = generate.filter(allDestinations, userPlaces, userDiff);
+    // Should return only 2nd destination
+    Set<Destination> expected = new HashSet();
+    expected.add(dest2);
```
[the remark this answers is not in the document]
shreyachatterjee00: @megankj Apparently this project runs on Java 8, and the ```.of``` methods can't be used until Java 9, and I'm not really sure how to change the version. I tried using the ```Set<String> stringSet = new HashSet<>(Arrays.asList("a", "b", "c"));``` syntax instead, and it gave me a bunch of errors (and is more difficult to read for me) so I figured I would just leave it for now. Is this okay or do you think there's a big benefit to adding all elements of the set together?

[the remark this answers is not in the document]
megankj: Ah drat.  Okay nevermind.

src/main/java/com/google/sps/data/Destination.java
```diff
@@ -48,6 +48,35 @@
...
+
+  public Obscurity getDifficulty() {
+    return this.level;
+  }
+
+  public String getCity() {
+    return this.city;
+  }
+
+  public static Obscurity stringToEnum(String difficulty) {
```
megankj, line 64: In the future if you want to add another static method that returns an enum, this name will be confusing.  Suggest renaming to toObscurity or something similar.

PR #39, opened 2020-08-07
Stores destinations created from user input in datastore

rysimone:


src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -44,10 +46,11 @@
   private static final String HINT3_PARAMETER = "hint3";
   private static final String OBSCURITY_PARAMETER = "obscurity";
   private static final String TAG_PARAMETER = "tag";
-  private static final String REDIRECT_URL = "/destination-data";
+  private static final String REDIRECT_URL = "/index.html";
```
lilymzhou, line 49: I might make this variable more specific (ex. "HOME_URL").

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -44,10 +46,11 @@
...
 
-  // Temporarily stores the destination created by the user
-  public List<Destination> destinations = new ArrayList<>();
```
shreyachatterjee00, line 50: If this is storing the destination that has just been created by the user, shouldn't this be just a Destination object instead of an array? Or if you're adding all the created Destinations to this array for now, maybe update this comment so it's more clear that it's all Destinations instead of just one

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
megankj: +1

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
rysimone: I was using that to store the destinations created during the prototype. I removed it because now we are storing in datastore.

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -44,10 +46,11 @@
...
+  private static final DatastoreService DATASTORE = DatastoreServiceFactory.getDatastoreService();
```
megankj, line 51: This isn't quite a constant since it's a service.  Don't use all caps in this case.

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -44,10 +46,11 @@
...
+
+  private static final Gson GSON = new Gson();
```
megankj, line 53: Same here, this probably shouldn't be capitalized.

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -92,21 +95,14 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
-  /* Retrieves the most recent destination created by the user, turns it into a JSON formatted string,
-   *and displays the JSON-ified String on /destination-data
-   */
-  @Override
-  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    Gson gson = new Gson();
-    String json = gson.toJson(destinations.get(destinations.size() - 1));
-    response.setContentType("application/json;");
-    response.getWriter().println(json);
+    Entity destinationEntity = new Entity("Destination");
```
megankj, line 101: Since I imagine this will be needed in multiple servlets, define this "Destination" string in Constants:
https://github.com/googleinterns/step183-2020/blob/master/src/main/java/com/google/sps/servlets/Constants.java

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -92,21 +95,14 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    destinationEntity.setProperty("value", jsonDestination);
```
megankj, line 102: Define this constant in Constats.java and also give it a more descriptive name, like destination_json or something.

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
tjgoog: On the other hand, I think that this is a constant according to the definition above. AFAICT, none of `Gson`'s methods have detectable side effects and its observable state cannot change. I think this name should stay as-is.

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -44,10 +46,11 @@
...
+  private static final String HOME_URL = "/index.html";
 
-  // Temporarily stores the destination created by the user
-  public List<Destination> destinations = new ArrayList<>();
+  private static final DatastoreService DATASTORE = DatastoreServiceFactory.getDatastoreService();
+
+  private static final Gson GSON = new Gson();
 
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
tjgoog, line 56: I think that this method should have its own Javadoc and that the class-level Javadoc should be updated now that the behavior is more fleshed out. I think that the class-level should be more general, like `Stores a destination in Datastore`, and the method-level should give a bit of detail in terms of its input and output, including: takes various parameters from the request, combines them to create a json string to represent a destination, stores the string in the datastore, and sends a redirect to HOME_URL.

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
megankj: I was trying to find a good explanation of why it shouldn't be in all caps but I was blanking out, so thanks for finding the passage. 👍

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
megankj: That's fine.  It seemed incorrect to me but I agree with you now.

PR #40, opened 2020-08-07
Add progress bar and timer; revise design flow of go.js.

lilymzhou:
- Added a progress bar that shows how far along the user is on the hunt.
- Added a timer that shows how long it has been since the user started the hunt.
- After beginning a new destination, the hint button is initially invisible, and shows up after a certain amount of time has passed.
- Revised the logical flow of go.js for clarity: added a ScavengerHuntManger class, revise handleDestinationAnswer() so that only one GET request is made to NameServlet.

Start:
Screenshot 2020-08-07 at 3 56 41 PM

Screenshot 2020-08-07 at 3 57 03 PM

Advancing through the hunt:
Screenshot 2020-08-07 at 3 57 33 PM

Final page:
Screenshot 2020-08-07 at 3 57 57 PM

### DIFF
--- a/src/main/webapp/go.js
+++ b/src/main/webapp/go.js
@@ -29,7 +29,7 @@ const SUBMIT_DISPLAY = 'response-area';
 const MAP_DISPLAY = 'map-area';
 const MAP_MSSG_DISPLAY = 'map-message-area';
 const TIMER_DISPLAY = 'timer-area';
-const GUESS_DISPLAY = 'guess-input';
+const GUESS_INPUT = 'guess-input';
 const PROGRESS_DISPLAY = 'progress-bar';
 
 // Hard-coded messages to be displayed to the user.
@@ -41,6 +41,8 @@ const WRONG_MSSG = 'Wrong. Try again!';
 const MAP_INTERVAL_MS = 10000; // ten seconds
 const HIDE_INTERVAL_MS = 5000; // five seconds
 // one second
+// Disabled lint check because TIMER_INTERVAL_MS is used
+// in scavengerHuntManager.js.
 const TIMER_INTERVAL_MS = 1000; // eslint-disable-line
 
 // Other constants.
@@ -50,6 +52,7 @@ const INVISIBLE_CLASS = 'invisible';
 // Global variables.
 let hunt;
 let map;
+let hintClock;
 
 window.onload = function() {
   addScriptToHead();
@@ -203,8 +206,8 @@ function handleDestinationAnswer() { //eslint-disable-line
   if (hunt.hasNotStarted()) { // User has not yet started the hunt.
     return;
   }
-  const userGuess = document.getElementById(GUESS_DISPLAY).value;
-  const queryStr = GUESS_URL + '?user-input=' + userGuess +
+  const userGuess = document.getElementById(GUESS_INPUT).value;
+  const queryStr = GUESS_URL + '?' + GUESS_INPUT + '=' + userGuess +
       '&answer=' + hunt.getCurDestName();
   fetch(queryStr).then((response) => response.json()).then((correctGuess) => {
     if (correctGuess) {
@@ -215,6 +218,7 @@ function handleDestinationAnswer() { //eslint-disable-line
           hunt.getCurDestDescription());
       addMarkerToMap(hunt.getCurDestLat(), hunt.getCurDestLng(),
           hunt.getCurDestName());
+      clearTimeout(hintClock);
     } else {
       updateMessage(SUBMIT_DISPLAY, WRONG_MSSG);
     }
@@ -247,6 +251,7 @@ function updateToCurrentState() {
   for (let i = 0; i < hunt.getDestIndex(); i++) {
     addMarkerToMap(hunt.getDest(i).lat, hunt.getDest(i).lng,
         hunt.getDest(i).name);
+  hunt.restartTimer();
   }
 }
 
@@ -340,7 +345,7 @@ function sendIndexToServlet(index) {
  * Disable lint check because proceed() is called from go.html.
  */
 function proceed() { //eslint-disable-line
-  hunt.destIndex = hunt.destIndex + 1;
+  hunt.incrementDestIndex();
   sendIndexToServlet(hunt.getDestIndex());
   if (hunt.getDestIndex() < hunt.getNumItems()) {
     updateMessage(RIDDLE_DISPLAY, 'Riddle: ' + hunt.getCurDestPuzzle());
@@ -372,7 +377,7 @@ function updateProgressBar() {
  */
 function delayHintButton() {
   toggleHintButton(/* hide = */ true);
-  window.setTimeout(toggleHintButton, HIDE_INTERVAL_MS, false);
+  hintClock = window.setTimeout(toggleHintButton, HIDE_INTERVAL_MS, false);
 }
 
 /**
--- a/src/main/webapp/go.js
+++ b/src/main/webapp/go.js
@@ -49,7 +49,6 @@ const INVISIBLE_CLASS = 'invisible';
 // Global variables.
 let hunt;
 let map;
-let startTime;
 
 window.onload = function() {
   addScriptToHead();
@@ -61,7 +60,7 @@ window.onload = function() {
  * and hours since the user started the scavenger hunt.
  */
 function updateTimer() {
-  const difference = new Date() - startTime;
+  const difference = hunt.getTimeElapsed();
   const seconds = difference % (1000 * 60) / 1000;
   const minutes = difference % (1000 * 60 * 60) / (1000 * 60);
   const hours = difference % (1000 * 60 * 60 * 24) /
@@ -110,7 +109,7 @@ function getHunt() {
           cur.location.lng));
     }
     hunt = new ScavengerHuntManager(destIndex, 0, huntArr);
-    updateToCurrentState(destIndex);
+    updateToCurrentState();
   });
 }
 
@@ -203,16 +202,16 @@ function handleDestinationAnswer() { //eslint-disable-line
   }
   const userGuess = document.getElementById(GUESS_DISPLAY).value;
   const queryStr = GUESS_URL + '?user-input=' + userGuess +
-      '&answer=' + hunt.getName();
+      '&answer=' + hunt.getCurDestName();
   fetch(queryStr).then((response) => response.json()).then((correctGuess) => {
     if (correctGuess) {
       toggleProceedButton(false);
       toggleHintButton(true);
       updateMessage(SUBMIT_DISPLAY, CORRECT_MSSG);
-      updateMessage(RIDDLE_DISPLAY, hunt.getName() + ': ' +
-          hunt.getDescription());
-      addMarkerToMap(hunt.getLat(), hunt.getLng(),
-          hunt.getName());
+      updateMessage(RIDDLE_DISPLAY, hunt.getCurDestName() + ': ' +
+          hunt.getCurDestDescription());
+      addMarkerToMap(hunt.getCurDestLat(), hunt.getCurDestLng(),
+          hunt.getCurDestName());
     } else {
       updateMessage(SUBMIT_DISPLAY, WRONG_MSSG);
     }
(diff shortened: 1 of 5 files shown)


src/main/webapp/go.js
```diff
@@ -150,30 +185,38 @@ function toggleHintButton(hide) { //eslint-disable-line
...
-  if (destIndex >= 0) {
-    const hintArr = huntArr[destIndex].hints;
-    if (hintIndex < hintArr.length) {
-      updateMessage(HINT_DISPLAY, hintArr[hintIndex]);
-      hintIndex++;
+  if (hunt.getDestIndex() >= 0) {
+    const hintArr = hunt.getCurHints();
+    if (hunt.getHintIndex() < hintArr.length) {
+      updateMessage(HINT_DISPLAY, hintArr[hunt.getHintIndex()]);
+      hunt.setHintIndex(hunt.getHintIndex() + 1);
```
shreyachatterjee00, line 192: Might be more cohesive to set hunt.getHintIndex() as another const variable, since you're getting it three times

src/main/webapp/go.js
```diff
@@ -150,30 +185,39 @@ function toggleHintButton(hide) { //eslint-disable-line
  * Disable lint check because getHint() is called from go.html.
  */
 function getHint() { //eslint-disable-line
-  if (destIndex >= 0) {
-    const hintArr = huntArr[destIndex].hints;
+  if (hunt.getDestIndex() >= 0) {
```
megankj, line 188: I'm thinking this logic could live inside the ScavengerHuntManager class.  You could call hunt.getNextHint() to get the next hint, if available and manage all the hintIndex logic.  The code here would then be much simpler and only in charge of updating the DOM:

```
function getHint() {
  const nextHint = hunt.getNextHint();
  if (nextHint != "") {
    updateMessage(HINT_DISPLAY, nextHint);
  }
}
```

src/main/webapp/go.js
```diff
@@ -150,30 +185,39 @@ function toggleHintButton(hide) { //eslint-disable-line
...
+ * Disable lint check because handleDestinationAnswer() is called
+ * from go.html.
  */
-function handleDestinationAnswer() {
-  fetch(GUESS_URL).then((response) => response.json()).then((guess) => {
-    if (guess === huntArr[destIndex].name) {
-      toggleProceedButton(/* hide = */ false);
-      toggleHintButton(/* hide = */ true);
+function handleDestinationAnswer() { //eslint-disable-line
+  if (hunt.getDestIndex() == -1) { // User has not yet started the hunt.
```
megankj, line 205: Consider writing a helper method to ScavengerHuntManager  make this more readable.  I.e. 

```
if (hunt.hasNotStarted()) {
  return;
}
```

src/main/webapp/go.js
```diff
@@ -150,30 +185,39 @@ function toggleHintButton(hide) { //eslint-disable-line
...
+    return;
+  }
+  const userGuess = document.getElementById(GUESS_DISPLAY).value;
+  const queryStr = GUESS_URL + '?user-input=' + userGuess +
+      '&answer=' + hunt.getCurName();
+  fetch(queryStr).then((response) => response.json()).then((result) => {
```
megankj, line 211: You can call this variable whatever you want.  Suggest calling it something like correctGuess so it's clear that a true value means they guessed correctly.
```suggestion
  fetch(queryStr).then((response) => response.json()).then((correctGuess) => {
```

src/main/webapp/go.js
```diff
@@ -150,30 +185,39 @@ function toggleHintButton(hide) { //eslint-disable-line
...
-      updateMessage(RIDDLE_DISPLAY, huntArr[destIndex].name + ': ' +
-          huntArr[destIndex].description);
-      addMarkerToMap(huntArr[destIndex].lat, huntArr[destIndex].lng,
-          huntArr[destIndex].name);
-    } else if (guess.length != 0) {
+      updateMessage(RIDDLE_DISPLAY, hunt.getCurName() + ': ' +
+          hunt.getCurDescription());
+      addMarkerToMap(hunt.getCurLat(), hunt.getCurLng(),
+          hunt.getCurName());
+    } else if (userGuess.length != 0) {
```
megankj, line 220: This only happens if they don't guess anything?  Shouldn't this just be an else?

src/main/webapp/go.js
```diff
@@ -184,9 +228,18 @@ function handleDestinationAnswer() {
  * Disable lint check because startHunt() is called from go.html.
  */
 function startHunt() { //eslint-disable-line
-  destIndex = 0;
+  hunt.incrementDestIndex();
```
megankj, line 231: Could just be hunt.start()

src/main/webapp/go.js
```diff
@@ -198,14 +251,15 @@ function updateToCurrentState(index) {
   if (index <= -1) { // The user has not yet pressed the start button.
     toggleHintButton(/* hide = */ true);
   } else { // The user has started the hunt, and needs to solve the riddle.
-    toggleHintButton(/* hide = */ false);
+    delayHintButton();
     hideStartButton();
-    updateMessage(RIDDLE_DISPLAY, 'Riddle: ' + huntArr[index].puzzle);
+    updateMessage(RIDDLE_DISPLAY, 'Riddle: ' + hunt.getDest(index).puzzle);
```
megankj, line 256: Will index be the same as destIndex which is available to the ScavengerHuntManager?  Because then it could just be hunt.getPuzzle() or something

src/main/webapp/go.js
```diff
@@ -231,7 +285,7 @@ function toggleProceedButton(hide) {
   if (hide) {
     proceedButton.classList.add(INVISIBLE_CLASS);
   } else {
-    if (destIndex === huntArr.length - 1) {
+    if (hunt.getDestIndex() === hunt.getNumItems() - 1) {
```
megankj, line 288: Could create a helper in ScavengerHuntManager to check this, i.e.

if (hunt.isAtLastStop())

src/main/webapp/go.js
```diff
@@ -150,30 +185,39 @@ function toggleHintButton(hide) { //eslint-disable-line
...
  * adjusts the display accordingly.
+ * Disable lint check because handleDestinationAnswer() is called
+ * from go.html.
  */
-function handleDestinationAnswer() {
-  fetch(GUESS_URL).then((response) => response.json()).then((guess) => {
-    if (guess === huntArr[destIndex].name) {
-      toggleProceedButton(/* hide = */ false);
-      toggleHintButton(/* hide = */ true);
+function handleDestinationAnswer() { //eslint-disable-line
```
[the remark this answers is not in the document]
lilymzhou: It seems like using the export keyword prevents go.js from executing -- I've tried exporting the function by changing the function declaration to `export function handleDestinationAnswer()`, as well as by adding the statement `export {handleDestinationAnswer}; `.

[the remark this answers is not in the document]
lilymzhou: Running this on the server doesn't cause any errors to arise, but the go.html stops handling button clicks, map insertion, etc.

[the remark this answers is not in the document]
megankj: Okay nevermind then!

src/main/webapp/go.js
lilymzhou: Yes! I initially added an else if to prevent the 'Wrong, try again!' method from displaying when the user doesn't type anything in, but it seems like handleDestinationAnswer() will only be called when the user hits the submit button.

src/main/webapp/go.js
```diff
@@ -150,30 +185,35 @@ function toggleHintButton(hide) { //eslint-disable-line
...
+    if (correctGuess) {
+      toggleProceedButton(false);
+      toggleHintButton(true);
       updateMessage(SUBMIT_DISPLAY, CORRECT_MSSG);
-      updateMessage(RIDDLE_DISPLAY, huntArr[destIndex].name + ': ' +
-          huntArr[destIndex].description);
-      addMarkerToMap(huntArr[destIndex].lat, huntArr[destIndex].lng,
-          huntArr[destIndex].name);
-    } else if (guess.length != 0) {
+      updateMessage(RIDDLE_DISPLAY, hunt.getName() + ': ' +
```
megankj, line 212: these all pertain to the current destination, so maybe a better name is getCurrentDestinationName(), etc.

OR have getCurrentDestination() return a Destination class, which then has getName() methods on it

src/main/webapp/go.js
```diff
@@ -198,14 +247,15 @@ function updateToCurrentState(index) {
   if (index <= -1) { // The user has not yet pressed the start button.
```
megankj, line 247: isn't index the same as hunt.destIndex?  You may not need to pass this variable around if hunt contains the updated index

src/main/webapp/go.js
```diff
@@ -184,9 +224,18 @@ function handleDestinationAnswer() {
  * Disable lint check because startHunt() is called from go.html.
  */
 function startHunt() { //eslint-disable-line
-  destIndex = 0;
+  hunt.start();
   sendIndexToServlet(0);
   updateToCurrentState(0);
+  startTimer();
```
megankj, line 230: I think the timer could also be moved to the hunt class as well, since when you start the hunt the timer should start.

The you could have methods like hunt.getTimeElapsed() which would return difference which you would use in updateTimer()

src/main/webapp/go.js
```diff
@@ -184,28 +225,27 @@ function handleDestinationAnswer() {
...
 
 /**
  * Updates the hunt to the current destination that the user is on.
- * @param {int} index The index of the destination that the user needs
- * to find.
  */
-function updateToCurrentState(index) {
-  if (index <= -1) { // The user has not yet pressed the start button.
+function updateToCurrentState() {
+  if (hunt.getDestIndex() <= -1) { // User has not pressed the start button.
```
megankj, line 238: You added a helper for this!
```suggestion
  if (hunt. hasNotStarted()) { // User has not pressed the start button.
```

src/main/webapp/go.js
```diff
@@ -299,19 +339,40 @@ function sendIndexToServlet(index) {
...
+  hunt.setHintIndex(0);
+}
+
+/**
+ * Update the progress bar to reflect how far the user is
+ * on the scavenger hunt.
+ */
+function updateProgressBar() {
+  const bar = document.getElementById(PROGRESS_DISPLAY);
+  const newWidth = (hunt.getDestIndex() / hunt.getNumItems()) * 100;
```
megankj, line 365: hunt could also have a method called getProgress that calculates this percentage

src/main/webapp/scavengerHuntManager.js
```diff
@@ -0,0 +1,140 @@
...
+   */
+  isAtLastStop() {
+    return this.destIndex === this.huntArr.length - 1;
+  }
+
+  /**
+   * Get the specific destination.
+   * @param {int} index Index of the destination
+   * to be retrieved.
+   * @return {String} Specific destination.
```
megankj, line 56: I think this returns a Destination not a string?

src/main/webapp/go.js
```diff
@@ -299,19 +340,39 @@ function sendIndexToServlet(index) {
  * Disable lint check because proceed() is called from go.html.
  */
 function proceed() { //eslint-disable-line
-  destIndex++;
-  sendIndexToServlet(destIndex);
-  if (destIndex < huntArr.length) {
-    updateMessage(RIDDLE_DISPLAY, 'Riddle: ' + huntArr[destIndex].puzzle);
-    toggleHintButton(/* hide = */ false);
+  hunt.destIndex = hunt.destIndex + 1;
```
tjgoog, line 343: Optional: I think it would be nice to bundle this up into a method on the manager class, e.g. `incrementDestIndex()`, or you could even bundle in the accessor part that happens on the next line with `incrementAndGetDestIndex()`.

src/main/webapp/go.js
```diff
@@ -299,19 +340,39 @@ function sendIndexToServlet(index) {
...
+}
+
+/**
+ * When the user first begins to find a destination, hide the hint
+ * button. Only show the hint button after a certain amount of time
+ * has passed.
+ */
+function delayHintButton() {
+  toggleHintButton(/* hide = */ true);
+  window.setTimeout(toggleHintButton, HIDE_INTERVAL_MS, false);
```
tjgoog, line 375: Does this need to be disabled once the user finds the current destination? Probably fine for now with a 5 second delay, but I think this should be a TODO in order to more cleanly handle the possibility of having longer delays. I imagine this situation could happen (let's pretend it's a 5 minute delay)
User starts scavenger hunt at t=0 minutes
User gets correct first destination and proceeds to next at t=4 minutes
User is trying to figure out the new destination, and at t=5 minutes, they get access to the hint button even though they have been looking for the current destination for only 1 minute.

src/main/webapp/scavengerHuntManager.js
```diff
@@ -0,0 +1,148 @@
...
+   * on the scavenger hunt.
+   */
+  constructor(destIndex, hintIndex, huntArr) {
+    this.destIndex = destIndex;
+    this.hintIndex = hintIndex;
+    this.huntArr = huntArr;
+  }
+
+  /**
+   * @return {String} Next hint that the user should see.
```
tjgoog, line 24: nit: Document that this also increments the hint index.

PR #41, opened 2020-08-11
Updates the destination creation form

rysimone:
Fixes latitude longitude input fields to accept numbers with up to 4 decimal places.
Changes the input type for the obscurity to radio so that only one can be chosen.

### DIFF
--- a/src/main/webapp/destinationCreation.html
+++ b/src/main/webapp/destinationCreation.html
@@ -13,9 +13,9 @@
         <input type="text" name="name" id="name-input" required>
         <p>Enter the latitude and longitude of the destination:</p>
         <label for="lat-input" class="required">Latitude:</label><br>
-        <input type="number" name="latitude" id="lat-input" required><br>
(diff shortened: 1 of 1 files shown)


src/main/webapp/destinationCreation.html
```diff
@@ -44,11 +44,11 @@ <h1>Create a Scavenger Hunt Destination</h1>
         <label for="family-input">Family</label><br>
         <br>
         <p class="required">Select how obscure this destination would be for tourists</p>
-        <input type="checkbox" name="obscurity" id="easy-input" value="easy">
+        <input type="radio" name="obscurity" id="easy-input" value="easy">
```
tjgoog, line 47: Could you add the `required` attribute to these now? https://stackoverflow.com/questions/8287779/how-to-use-the-required-attribute-with-a-radio-input-field

Looks like putting `required` in any one of the inputs will make it so that one option in the entire group must be selected, but as mentioned in that link and the corresponding standard, it's a best practice to put `required` in all of them so it's more clear when someone's editing this file.

PR #43, opened 2020-08-12
Retrieve scavenger hunt data from Datastore in GoDataServlet.java

lilymzhou:
- HuntID is sent to go.html, which is then sent as a parameter to GoDataServlet. GoDataServlet retrieves the relevant scavenger hunt from Datastore.
- Added an updateIndex() method to ScavengerHunt.java, and removed the city field, which is unused.

Assumes data is stored in the following format (Entity = "ScavengerHunt", Property = "ScavengerHunt")
Screenshot 2020-08-13 at 10 27 10 AM

HuntID passed from filter.js in the format: "huntID=[huntEntity.getKey()]"
Screenshot 2020-08-12 at 12 12 19 PM


src/main/java/com/google/sps/servlets/GoDataServlet.java
```diff
@@ -14,108 +14,84 @@
...
+
+      // Update index of scavenger hunt.
+      ScavengerHunt hunt =
+          gson.fromJson((String) huntEntity.getProperty(Constants.HUNT_VAL), ScavengerHunt.class);
+      hunt.updateIndex(index);
+
+      // Convert hunt to JSON string and put into Datastore
+      String huntStr = gson.toJson(hunt);
+      huntEntity.setProperty(Constants.HUNT_VAL, huntStr);
+      datastore.put(huntEntity);
```
megankj, line 61: The main purpose of this code is to update the index, but in order to do so you have to modify the whole hunt_val string, which isn't ideal.

I think in an alternate design hunt_val would be independent of hunt_index so you could update them independently without having to convert back and forth.  I will not make this blocking, but I just wanted to note it here.

src/main/java/com/google/sps/servlets/GoDataServlet.java
```diff
@@ -14,108 +14,84 @@
...
-    items.add(secondHunt);
-    items.add(thirdHunt);
-    ScavengerHunt hunt = new ScavengerHunt(items, index, "San Francisco");
-    return hunt;
+  /**
+   * Retrieves the scavenger hunt from Datastore using {@code huntID}, the ID corresponding to the
+   * scavenger hunt that should be retrieved.
+   */
+  private Entity findScavengerHunt(String huntID) {
+    Query huntQuery = new Query(Constants.SCAVENGER_HUNT_ENTITY);
```
megankj, line 88: Instead of retrieving all entities and then filtering for the key, can you set up the Query to already be looking for this key?

https://cloud.google.com/datastore/docs/concepts/queries#key_filters

src/main/webapp/go.js
```diff
@@ -105,7 +108,11 @@ function addScriptToHead() {
  * to reflect the current state of the hunt.
  */
 function getHunt() {
-  fetch(DATA_URL).then((response) => response.json()).then((mssg) => {
+  // ID of the scavenger hunt that the user is on,
+  // in the form of: hunt_id=ScavengerHunt([ID number here])
+  huntID = window.location.search.substr(1);
```
megankj, line 113: You should note that you're using substr(1) here because the first character in window.location.search is '?'

src/main/webapp/go.js
```diff
@@ -105,7 +108,11 @@ function addScriptToHead() {
  * to reflect the current state of the hunt.
  */
 function getHunt() {
-  fetch(DATA_URL).then((response) => response.json()).then((mssg) => {
+  // ID of the scavenger hunt that the user is on,
+  // in the form of: hunt_id=ScavengerHunt([ID number here])
```
megankj, line 112: Does this mean the URL literally looks like hunt_id=ScavengerHunt(123)?  Why do you need the ScavengerHunt part?  Can this just be hunt_id=123?

src/main/webapp/go.js
```diff
@@ -328,13 +335,16 @@ function updateRiddleToFinalMessage() {
...
+ * the user is on. Also sends back the ID of the scavenger hunt
+ * that the user is on.
  * @param {int} index Index of the current destination the user needs
  * to find.
  */
 function sendIndexToServlet(index) {
   const params = new URLSearchParams();
   params.append(INDEX_PARAM, index);
+  const pair = huntID.split('=');
+  params.append(pair[0], pair[1]);
```
megankj, line 347: You always need to check values before you access them.  check if pair.length >=2

src/main/webapp/go.js
lilymzhou: This is mostly for simplicity; I believe that entity.getKey() returns the string "ScavengerHunt([some number here])."

megankj: But then won't GoDataServlet try to parse ScavengerHunt(123) as a number?  Won't that not return what you expect?

lilymzhou: GoDataServlet currently tries to parse ScavengerHunt([number]) as a string, which is used to find the corresponding entity in Datastore.

megankj: Oh I see, my mistake.  I think it would be better to just pass around the ID, which you could then use as a filter when you are querying datastore.  But for the sake of simplicity, I think this is fine.

lilymzhou: That makes sense -- I actually just changed huntID to a long instead of a String in the latest commit; turns out that it's easier to efficiently retrieve an entity from Datastore by key using the ID number.

src/main/java/com/google/sps/servlets/GoDataServlet.java
lilymzhou: If I stored hunt_index independently from hunt_val, I think I'd still need to modify the entire hunt_val string in the doGet method, where I send the scavenger hunt to go.js, based on my current design.

Another possibility could be changing the ScavengerHunt class to not include the index, though that'd involve figuring out a way to send both the index and the scavenger hunt in the doGet() response.

megankj: We chatted about this and brainstormed a way to do this and decided to focus on it next week.  thanks!

src/main/java/com/google/sps/servlets/GoDataServlet.java
```diff
@@ -14,108 +14,89 @@
...
+    String huntIDStr = request.getParameter(Constants.HUNTID_PARAMETER);
+    Entity huntEntity = null;
+    try {
+      long huntID = Long.parseLong(huntIDStr);
+      huntEntity = findScavengerHunt(huntID);
+    } catch (Exception e) {
+    }
+
+    if (huntEntity == null) {
+      return;
```
megankj, line 82: what happens on the client side if the huntEntity is null?  The user might want to see an error message or something, right?

src/main/java/com/google/sps/servlets/GoDataServlet.java
```diff
@@ -14,108 +14,89 @@
...
-    items.add(secondHunt);
-    items.add(thirdHunt);
-    ScavengerHunt hunt = new ScavengerHunt(items, index, "San Francisco");
-    return hunt;
+  /**
+   * Retrieves the scavenger hunt from Datastore using {@code huntID}, the ID corresponding to the
+   * scavenger hunt that should be retrieved.
+   */
+  private Entity findScavengerHunt(long huntID) {
+    Key key = KeyFactory.createKey(Constants.SCAVENGER_HUNT_ENTITY, huntID);
```
megankj, line 95: nice!

lilymzhou:
I just rebased this branch, which seemed to have added the contents of Shreya's recent PR to this PR -- the changes unique to this branch are in ScavengerHunt.java, Constants.java, GoDataServlet.java, and go.js.

src/main/java/com/google/sps/servlets/GoDataServlet.java
lilymzhou: I just set the response to an error message within this if loop.

src/main/webapp/go.js
```diff
@@ -105,7 +108,12 @@ function addScriptToHead() {
  * to reflect the current state of the hunt.
  */
 function getHunt() {
-  fetch(DATA_URL).then((response) => response.json()).then((mssg) => {
+  // ID of the scavenger hunt that the user is on,
+  // in the form of: hunt_id=ScavengerHunt([ID number here])
+  // Using index 1 because index 0 is the '?' character.
+  huntID = window.location.search.substr(1);
```
tjgoog, line 114: Optional:
```suggestion
  huntID = new URLSearchParams(window.location.search).get('hunt_id');
```
is an option for a more structured way to get the value. Though then you'd have to explicitly put the `hunt_id=` part in to the `queryStr`.

PR #44, opened 2020-08-12
Push Scavenger Hunt objects to Datastore

shreyachatterjee00:
* Convert destination objects to Hunt Items 
* Create Scavenger Hunt 
* Store Scavenger Hunt in Datastore 

* What a Scavenger Hunt object looks like in Datastore: 
Screenshot 2020-08-12 at 4 39 17 PM

* In JS file: 
* if Error message is returned, display div that shows error message
image

* If Id is returned, redirect to go.html with the hunt Id

### DIFF
--- a/src/main/java/com/google/sps/data/Destination.java
+++ b/src/main/java/com/google/sps/data/Destination.java
@@ -70,7 +70,7 @@ public class Destination {
   }
 
   /* Return one Riddle. TODO: Make this a random riddle. */
-  public Riddle getRiddle() {
+  public Riddle getRandomRiddle() {
     return riddles.get(0);
   }
 
@@ -96,7 +96,7 @@ public class Destination {
             .withName(this.name)
             .atLocation(this.location)
             .withDescription(this.description)
-            .withRiddle(this.riddles.get(0))
+            .withRiddle(this.getRiddle())
             .build();
 
     return item;
--- a/src/main/java/com/google/sps/data/Destination.java
+++ b/src/main/java/com/google/sps/data/Destination.java
@@ -102,7 +102,6 @@ public class Destination {
     return item;
   }
 
-
   public static class Builder {
     private String name;
     private LatLng location;
--- a/src/main/java/com/google/sps/data/Destination.java
+++ b/src/main/java/com/google/sps/data/Destination.java
@@ -61,6 +61,19 @@ public class Destination {
     return this.city;
   }
 
+  public String getDescription() {
+    return this.description;
+  }
+
+  public LatLng getLocation() {
+    return this.location;
+  }
+
+  /* Return one Riddle. TODO: Make this a random riddle. */
+  public Riddle getRiddle() {
+    return riddles.get(0);
+  }
+
   public static Obscurity stringToObscurity(String difficulty) {
     Obscurity level = Obscurity.UNDEFINED;
     switch (difficulty.toLowerCase()) {
@@ -76,6 +89,19 @@ public class Destination {
     }
     return level;
   }
+
+  public HuntItem convertToHuntItem() {
+    HuntItem item =
+        new HuntItem.Builder()
+            .withName(this.name)
+            .atLocation(this.location)
+            .withDescription(this.description)
+            .withRiddle(this.riddles.get(0))
+            .build();
+
+    return item;
+  }
+
 
   public static class Builder {
     private String name;
--- a/src/main/java/com/google/sps/data/Destination.java
+++ b/src/main/java/com/google/sps/data/Destination.java
@@ -96,7 +96,7 @@ public class Destination {
             .withName(this.name)
             .atLocation(this.location)
             .withDescription(this.description)
-            .withRiddle(this.getRiddle())
+            .withRiddle(this.getRandomRiddle())
             .build();
 
     return item;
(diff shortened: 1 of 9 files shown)


src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -58,10 +67,15 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     Set<Destination> filteredDestinations =
         filter(allDestinations, userPlaces, userDifficultyLevels);
 
-    writeToDataStore(filteredDestinations);
+    // Convert Destinations to Hunt Items, create Scavenger Hunt, store in Datastore
+    // TODO: Only return the amount of hunt items that the user wants
+    ArrayList<HuntItem> huntItems = convertToHuntItems(filteredDestinations);
+    ScavengerHunt scavHunt = new ScavengerHunt(huntItems, -1);
```
megankj, line 73: You can make another constructor for ScavengerHunt  that only takes huntItems.  In that case, you can assume -1 for the index.

Side note: at some point we should probably discuss the idea of using index = 0 instead of -1.  I know a lot of the logic in go.js would need to change, but using -1 is not intuitive, since this should ideally just be a pointer to the "active" hunt item.

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -79,54 +93,89 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
-  public void writeToDataStore(Set<Destination> filteredDestinations) {
-    System.out.println(filteredDestinations);
+  /* Convert all Destinations in set to Hunt Items. */
+  public ArrayList<HuntItem> convertToHuntItems(Set<Destination> allDestinations) {
+    ArrayList<HuntItem> filteredHuntItems = new ArrayList();
+    for (Destination destination : allDestinations) {
+      HuntItem item = destination.convertToHuntItem();
+      filteredHuntItems.add(item);
+    }
+    return filteredHuntItems;
```
megankj, line 103: You don't need to change it if you prefer the for loop but you can eliminate this method and do this logic in one line with my all time fav, streams!

`allDestinations.stream().map(Destination::convertToHuntItem).collect(Collectors.toList())`

The `map(Destination::convertToHuntItem)` part is equivalent to `map(destination -> destination.convertToHuntItem())`

src/main/webapp/filter.js
```diff
@@ -62,8 +62,17 @@ function sendClickedFiltersToServer() { //eslint-disable-line
     DIFF_ARRAY_URL_PARAM + '=' + jsonDiffArray,
   {method: 'POST'}).then((response) => response.text())
       .then((message) => {
-        // TODO: take out, replace with success or error message
-        console.log(message);
+        // If success, redirect to scavenger hunt with ID
+        // If failure, display on screen
+        if (message != 'Error\n') {
```
megankj, line 67: Could you use trim here?
https://www.w3schools.com/jsref/jsref_trim_string.asp

megankj: It might also be better to return a json object with a success boolean and a hunt_id field if it was successful, but this doesn't seem to be causing any issues at the moment, so it's okay

src/main/java/com/google/sps/data/Destination.java
```diff
@@ -61,6 +61,19 @@ public String getCity() {
...
+    return this.description;
+  }
+
+  public LatLng getLocation() {
+    return this.location;
+  }
+
+  /* Return one Riddle. TODO: Make this a random riddle. */
+  public Riddle getRiddle() {
+    return riddles.get(0);
```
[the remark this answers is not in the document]
shreyachatterjee00: got it!

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -58,10 +67,15 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     Set<Destination> filteredDestinations =
         filter(allDestinations, userPlaces, userDifficultyLevels);
 
-    writeToDataStore(filteredDestinations);
+    // Convert Destinations to Hunt Items, create Scavenger Hunt, store in Datastore
+    // TODO: Only return the amount of hunt items that the user wants
```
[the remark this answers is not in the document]
shreyachatterjee00: That makes more sense, yeah! I'll move the TODO

src/main/java/com/google/sps/data/Destination.java
```diff
@@ -61,6 +61,19 @@ public String getCity() {
...
+  public String getDescription() {
+    return this.description;
+  }
+
+  public LatLng getLocation() {
+    return this.location;
+  }
+
+  /* Return one Riddle. TODO: Make this a random riddle. */
+  public Riddle getRiddle() {
```
tjgoog, line 73: I think this would be better as `getRandomRiddle()`, but that change can wait until you actually make it random.

src/main/java/com/google/sps/data/Destination.java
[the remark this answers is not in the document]
tjgoog: To expand: It's OK to access hard-coded indexes if you have already done validation that the index does exist. e.g. wrapping the access in an `if (size > blah)`, or validating the data when the object is constructed if the class is set up in a way such that the data can't change after construction.

PR #47, opened 2020-08-14
Get destinations from Datastore

shreyachatterjee00:
* Transition from getting Destination objects from Datastore instead of creating fake data & return Scavenger Hunt Id

* In datastore after creating three Destinations through @rysimone 's form: 
Screenshot 2020-08-14 at 10 29 13 AM

* In datastore after creating a Scavenger Hunt (filtered by SF and Easy): 
Screenshot 2020-08-14 at 10 30 01 AM

* Redirect to @lilymzhou 's page: 
Screenshot 2020-08-14 at 10 29 45 AM


src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -104,87 +119,12 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
   }
 
   /* Store Scavenger Hunt object in Datastore. */
```
tjgoog, line 121: nit: Mention in javadoc that the return value is the key of the new datastore entry.

PR #49, opened 2020-08-17
Add index as a separate Scavenger Hunt property in Datastore.

lilymzhou:
ScavengerHunt entities in Datastore:
Screenshot 2020-08-17 at 10 24 50 AM

### DIFF
--- a/src/main/java/com/google/sps/servlets/GoDataServlet.java
+++ b/src/main/java/com/google/sps/servlets/GoDataServlet.java
@@ -82,7 +82,9 @@ public class GoDataServlet extends HttpServlet {
     } else {
       ScavengerHunt hunt =
           gson.fromJson((String) huntEntity.getProperty(Constants.HUNT_VAL), ScavengerHunt.class);
-      ScavengerHuntFull fullHunt = new ScavengerHuntFull(hunt, ((Long) huntEntity.getProperty(Constants.HUNT_INDEX)).intValue());
+      ScavengerHuntFull fullHunt =
+          new ScavengerHuntFull(
+              hunt, ((Long) huntEntity.getProperty(Constants.HUNT_INDEX)).intValue());
       String json = gson.toJson(fullHunt);
       response.getWriter().println(json);
     }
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/sps/servlets/GoDataServlet.java
```diff
@@ -54,13 +53,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
 
       // Update index of scavenger hunt.
-      ScavengerHunt hunt =
-          gson.fromJson((String) huntEntity.getProperty(Constants.HUNT_VAL), ScavengerHunt.class);
-      hunt.updateIndex(index);
-
-      // Convert hunt to JSON string and put into Datastore
-      String huntStr = gson.toJson(hunt);
-      huntEntity.setProperty(Constants.HUNT_VAL, huntStr);
+      huntEntity.setProperty(Constants.HUNT_INDEX, index);
```
megankj, line 56: I think you'll only want to do this if huntEntity != null

src/main/java/com/google/sps/servlets/GoDataServlet.java
```diff
@@ -84,7 +77,9 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     if (huntEntity == null) {
       response.getWriter().println(ERROR_MSSG);
     } else {
-      String json = (String) huntEntity.getProperty(Constants.HUNT_VAL);
+      String jsonHunt = (String) huntEntity.getProperty(Constants.HUNT_VAL);
+      int curIndex = ((Long) huntEntity.getProperty(Constants.HUNT_INDEX)).intValue();
+      String json = jsonHunt.substring(0, jsonHunt.length() - 1) + ",\"index\":" + curIndex + "}";
```
megankj, line 82: Instead of modifying the string itself (which may be error prone), I think you should construct the JSON using a library of some kind (for example, [JsonObjectBuilder](https://docs.oracle.com/javaee/7/api/javax/json/JsonObjectBuilder.html)).

```
{
  hunt: {
    items: {
    }
  },
  index: -1
}
```
It will add a level of nesting, but also gives you the flexibility to add other fields in the future if you want/need to.

PR #50, opened 2020-08-17
Autogenerate hints

lilymzhou:
- Implementation autogeneration of hints using Places API.
  - All instances where the user review directly contains the destination name is replaced with "[REDACTED]."
- Add pop-up message when user clicks the exit button that reminds them to save or bookmark the link if they want to return to the hunt later.
- Selectively display the user input form.
- Bold the riddle for readability, and reformat the way that hints are displayed to account for the increase in height of the display.

Start of hunt:
Screenshot 2020-08-17 at 11 47 41 AM

Autogenerating hints (reviews):
Screenshot 2020-08-17 at 11 49 53 AM

Autogenerating hints (photos):
Screenshot 2020-08-17 at 11 52 39 AM

After completing the hunt:
Screenshot 2020-08-17 at 12 03 56 PM

### DIFF
--- a/src/main/webapp/go.html
+++ b/src/main/webapp/go.html
@@ -32,9 +32,10 @@
           <button class="button" id="start-button" type="button" onclick="startHunt()">Start Hunt</button>
           <button class="button" id="hint-button" type="button" onclick="getHint()">Get a Hint</button>
           <button class="button" id="proceed-button" type="button" onclick="proceed()">Proceed to Next Destination</button>
+          <button class="button" id="clear-hint-button" type="button" onclick="deleteMessage(HINT_DISPLAY)">Clear Hints</button>
         </div>
-        <div class="center-block hint-row" id="hint-area"></div>
       </div>
+      <div class="center-block hint-row" id="hint-area"></div>
       <div class="footer">
         <button class="button" type="button" onclick="exit()">Exit</button>
       </div>
--- a/src/main/webapp/go.js
+++ b/src/main/webapp/go.js
@@ -307,7 +307,7 @@ function displayNextReview() {
   const curIndex = hunt.getReviewIndex();
   if (curIndex < hunt.getReviews().length) {
     updateMessage(HINT_DISPLAY, 'Review of destination: ');
-    let cleanedReview = redactReview(hunt.getReviews()[curIndex].text);
+    const cleanedReview = redactReview(hunt.getReviews()[curIndex].text);
     updateMessage(HINT_DISPLAY, cleanedReview);
     hunt.setReviewIndex(curIndex + 1);
   }
@@ -319,7 +319,7 @@ function displayNextReview() {
  * @return {String} Redacted text
  */
 function redactReview(text) {
-  const re = new RegExp(hunt.getCurDestName(), "gi")
+  const re = new RegExp(hunt.getCurDestName(), 'gi');
   return text.replace(re, '[REDACTED]');
 }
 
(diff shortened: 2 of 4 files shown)


src/main/webapp/go.html
```diff
@@ -32,9 +32,10 @@
           <button class="button" id="start-button" type="button" onclick="startHunt()">Start Hunt</button>
           <button class="button" id="hint-button" type="button" onclick="getHint()">Get a Hint</button>
           <button class="button" id="proceed-button" type="button" onclick="proceed()">Proceed to Next Destination</button>
+          <button class="button" id="clear-hint-button" type="button" onclick="deleteMessage(HINT_DISPLAY)">Clear Hints</button>
```
tjgoog, line 35: Does `deleteMessage(HINT_DISPLAY)` work as expected? I'm not sure, because `HINT_DISPLAY` may not be referenceable from this location, but perhaps it is.

src/main/webapp/go.js
```diff
@@ -200,10 +206,135 @@ function toggleHintButton(hide) { //eslint-disable-line
 function getHint() { //eslint-disable-line
   const nextHint = hunt.getNextHint();
   if (nextHint != '') {
-    updateMessage(HINT_DISPLAY, nextHint);
+    updateMessage(HINT_DISPLAY, 'Hint #' + (hunt.getHintIndex()) +
+        ': ' + nextHint);
+  } else {
+    if (hunt.getPlaceID() == -1) {
+      autoGenerateHints();
```
tjgoog, line 213: Why is `autoGenerateHints()` only called if the user hasn't started the hunt yet? Is it possible that this method won't be called before the user beings, and if so, will this cause problems for `getNextAutoHint()`, e.g. the auto hints might not exist?

tjgoog: Update: Maybe placeID is the Google Places API ID? If this code is correct, could you add some comments to explain what it's doing?

tjgoog: Update 2: It might be cleaner to move this branching logic and into `getNextAutoHint()`. Here's a pseudocode version of what I'm thinking:

```
function getNextAutoHint() {
  if (hunt.getPlaceID() == -1) {
    autoGenerateHints();    
  } else {
    const random = Math.random();
    // And so on...
  }
}
```

src/main/webapp/scavengerHuntManager.js
```diff
@@ -44,6 +49,62 @@ class ScavengerHuntManager { //eslint-disable-line
...
+   * This method is called when the user moves on to a new
+   * destination.
+   */
+  resetForNextDest() {
+    hunt.setHintIndex(0);
+    hunt.setPhotoIndex(0);
+    hunt.setReviewIndex(0);
+    hunt.setPhotos([]);
+    hunt.setReviews([]);
+    hunt.setPlaceID(-1);
```
tjgoog, line 105:
```suggestion
  resetForNextDest() {
    this.setHintIndex(0);
    this.setPhotoIndex(0);
    this.setReviewIndex(0);
    this.setPhotos([]);
    this.setReviews([]);
    this.setPlaceID(-1);
```

src/main/webapp/go.js
lilymzhou: Yes, getPlaceID() refers to the unique ID for a destination provided by the Places API -- I added comments to clarify this, and restructured the branching logic.

src/main/webapp/go.html
lilymzhou: yes, this works as expected!

PR #54, opened 2020-08-18
User enters number of Destinations they want, between 1 and 10

shreyachatterjee00:
* User can click either 3 or 5 Destinations
* If both 3 and 5 are clicked, alert says to pick one: 
Screenshot 2020-08-18 at 3 35 11 PM

* If neither are picked, 3 is default
* If there are not enough Destinations to match the user's filters, an error will be shown: 
Screenshot 2020-08-18 at 3 35 02 PM

* Otherwise, redirects to going on the Scavenger Hunt 

Update 08/19: 
* User can input a number for the number of destinations they want
Screenshot 2020-08-19 at 4 55 26 PM

### DIFF
--- a/src/main/java/com/google/sps/servlets/GenerateServlet.java
+++ b/src/main/java/com/google/sps/servlets/GenerateServlet.java
@@ -124,8 +124,8 @@ public class GenerateServlet extends HttpServlet {
   }
 
   /** 
-  * Returns the filteredDestinations Set with the correct number of Destinations. 
-  * Chooses a random subset of the filtered Destinations.
+  * Returns the filteredDestinations Set with the correct number of Destinations. Chooses a random
+  * subset of the filtered Destinations.
   **/
   public Set<Destination> correctNumDests(Set<Destination> filteredDestinations, int numPlaces) {
     if (filteredDestinations.size() == numPlaces) {
--- a/src/main/java/com/google/sps/servlets/GenerateServlet.java
+++ b/src/main/java/com/google/sps/servlets/GenerateServlet.java
@@ -24,7 +24,9 @@ import com.google.sps.data.Destination;
 import com.google.sps.data.HuntItem;
 import com.google.sps.data.ScavengerHunt;
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -33,8 +35,6 @@ import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-import java.util.Collections;
-import java.util.Arrays;
 
 /** Servlet that returns bucket list content */
 @WebServlet("/generate-hunt")
@@ -53,8 +53,7 @@ public class GenerateServlet extends HttpServlet {
         new Gson().fromJson(request.getParameter(PLACE_FILTERS), HashSet.class);
     HashSet<String> userDifficultyStrings =
         new Gson().fromJson(request.getParameter(DIFF_FILTERS), HashSet.class);
-    String numPlacesString = 
-        new Gson().fromJson(request.getParameter(NUM_PLACES), String.class);
+    String numPlacesString = new Gson().fromJson(request.getParameter(NUM_PLACES), String.class);
     int numPlaces = 3;
     if (numPlacesString == "Five") {
       numPlaces = 5;
@@ -81,7 +80,7 @@ public class GenerateServlet extends HttpServlet {
     if (filteredDestinations.size() >= numPlaces) {
       // Get numPlaces number of Destinations
       filteredDestinations = correctNumDests(filteredDestinations, numPlaces);
-     
+
       // Convert Destinations to Hunt Items, create Scavenger Hunt, store in Datastore
       ArrayList<HuntItem> huntItems = convertToHuntItems(filteredDestinations);
       ScavengerHunt scavHunt = new ScavengerHunt(huntItems);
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -50,6 +53,11 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
         new Gson().fromJson(request.getParameter(PLACE_FILTERS), HashSet.class);
     HashSet<String> userDifficultyStrings =
         new Gson().fromJson(request.getParameter(DIFF_FILTERS), HashSet.class);
+    String numPlacesString = new Gson().fromJson(request.getParameter(NUM_PLACES), String.class);
+    int numPlaces = 3;
+    if (numPlacesString == "Five") {
```
lilymzhou, line 58: I think .equals() is usually used in Java for String comparison (rather than ==).

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -108,6 +126,24 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    if (filteredDestinations.size() == numPlaces) {
+      return filteredDestinations;
+    }
+    Set<Destination> newFilteredDests = new HashSet<Destination>();
+    List<Destination> filteredList = new ArrayList<Destination>(filteredDestinations);
+    Integer[] indexArr = new Integer[filteredDestinations.size()];
+    for (int i = 0; i < indexArr.length; i++) {
+      indexArr[i] = i;
+    }
+    Collections.shuffle(Arrays.asList(indexArr));
```
lilymzhou, line 140: wow I didn't know that a function like this existed! this is pretty cool.

src/main/java/com/google/sps/servlets/GenerateServlet.java
shreyachatterjee00: Oh you're totally right, I forgot!

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -108,6 +126,24 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     return filteredDestinations;
   }
 
+  /* Returns the filteredDestinations Set with the correct number of Destinations. */
```
tjgoog, line 129: Could you also mention here that it chooses a random subset of destinations to meet that filtering requirement?

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -108,6 +126,24 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+  public Set<Destination> correctNumDests(Set<Destination> filteredDestinations, int numPlaces) {
+    if (filteredDestinations.size() == numPlaces) {
```
tjgoog, line 131: Do you think this would be better as `<=` in order to handle situations where there are fewer destinations than the requested amount?

src/main/webapp/filter.js
```diff
@@ -55,11 +56,29 @@ function sendClickedFiltersToServer() { //eslint-disable-line
...
+  // Get clicked num stops
+  let clickedNumStops = [];
+  let numStops = '';
+  if (document.getElementsByClassName('num-stops').length > 0) {
+    const numStopsContainer = document.getElementsByClassName('num-stops')[0];
+    clickedNumStops =
+        Array.from(numStopsContainer.getElementsByClassName(CLICKED))
+            .map((element) => element.innerText);
+    if (clickedNumStops.length > 1) {
+      window.alert('Please choose only ONE number of stops.');
```
tjgoog, line 68: When this happens, should the method return so that the request to the server is not made?

megankj: I think that you should enforce this in the UI (i.e. only allow them to click one of the numbers, not both 3 and 5) instead of letting them click anything and then triggering an alert.

I'm also curious why it's only 3 or 5?  The number seems arbitrary.  Why can't they select any number?

src/main/java/com/google/sps/servlets/GenerateServlet.java
shreyachatterjee00: I handle the case where there are fewer Destinations than the requested amount in the else statement in the doPost function, because if this is the case, I can just immediately return an error. So, in this function, I already know that there are enough filteredDestinations.

tjgoog: 👍

src/main/java/com/google/sps/servlets/GenerateServlet.java
```diff
@@ -64,18 +69,28 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    // If there are enough destinations to return, pick random ones to put in Hunt
+    if (filteredDestinations.size() >= numPlaces) {
+      // Get numPlaces number of Destinations
+      filteredDestinations = correctNumDests(filteredDestinations, numPlaces);
+
+      // Convert Destinations to Hunt Items, create Scavenger Hunt, store in Datastore
+      ArrayList<HuntItem> huntItems = convertToHuntItems(filteredDestinations);
+      ScavengerHunt scavHunt = new ScavengerHunt(huntItems);
+      long huntId = writeToDataStore(scavHunt);
+      System.out.println(huntId);
```
megankj, line 84: delete this?

src/main/webapp/filter.js
```diff
@@ -55,11 +56,16 @@ function sendClickedFiltersToServer() { //eslint-disable-line
   }
   const jsonDiffArray = JSON.stringify(clickedDifficulties);
 
-  // TODO: Get clicked num stops
+  // Get clicked num stops
+  numStops = document.getElementById('num-stops').value;
+  console.log(numStops);
```
megankj, line 61: delete?

src/main/webapp/filter.js
```diff
@@ -55,11 +56,16 @@ function sendClickedFiltersToServer() { //eslint-disable-line
...
+  const jsonNumPlaces = JSON.stringify(numStops);
```
megankj, line 62: I don't think this needs to be JSON formatted (JSON.stringify(3) returns "3").

PR #56, opened 2020-08-19
Create clickable divs from the place objects returned from the google place query

rysimone:
When a div is clicked, the name, latitude, and longitude values for the place are inputted into their appropriate fields.
The map also re-centers itself around the place that has been selected and drops a marker at its location.

image

### DIFF
--- a/src/main/webapp/destinationCreation.js
+++ b/src/main/webapp/destinationCreation.js
@@ -36,17 +36,28 @@ function searchForPlace() { // eslint-disable-line
 
   const request = {
     query: text,
-    fields: ['name', 'geometry'],
+    fields: ['name', 'geometry', 'place_id', 'icon', 'photos'],
   };
 
   placeService.findPlaceFromQuery(request, (results, status) => {
+    let newDiv;
     if (status === google.maps.places.PlacesServiceStatus.OK) {
-      results.forEach((item) => {
-        // TODO: Create clickable divs for each item that is
-        // returned from the query and fill in input fields with appropriate
-        // information provided by the places object
-        console.log(item);
+      results.forEach((place) => {
+        newDiv = document.createElement('div');
+        newDiv.style = 'cursor: pointer;';
+        newDiv.setAttribute('place-id', place.place_id);
+        newDiv.innerText = place.name;
+        newDiv.id = 'place';
+        document.getElementsByClassName('search-results')[0].appendChild(newDiv);
+        //newImg = document.createElement('img');
+        //newImg.src = place.photos[0].getUrl();
+        //document.getElementById(place.name)[0].appendChild(newImg);
+        console.log(place);
       });
     }
   });
+}
+
+function fillInValues() { // eslint-disable-line
+  console.log('Fill me in');
 }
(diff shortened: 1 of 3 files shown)


src/main/webapp/destinationCreation.js
```diff
@@ -22,6 +22,7 @@ function addScriptToHead() { // eslint-disable-line
 }
 
 function searchForPlace() { // eslint-disable-line
+  document.getElementsByClassName('search-results').innerHtml = '';
```
lilymzhou, line 25: I might make 'search-results' a const, since this is used more than once in this file.

rysimone: Yea you are right! thanks

src/main/webapp/destinationCreation.js
```diff
@@ -22,31 +32,87 @@ function addScriptToHead() { // eslint-disable-line
 }
 
 function searchForPlace() { // eslint-disable-line
+  const element = document.querySelectorAll('#place');
```
megankj, line 35: IDs are supposed to be unique, and since you want to have multiple divs, you should probably be using class to query for the elements, not ID

src/main/webapp/destinationCreation.js
```diff
@@ -22,31 +32,87 @@ function addScriptToHead() { // eslint-disable-line
...
+  if (element.length > 0) {
+    let i;
+    for (i = 0; i < element.length; i++) {
```
megankj, line 38: You don't need to define i before the for loop
```suggestion
    for (let i = 0; i < element.length; i++) {
```

src/main/webapp/destinationCreation.js
```diff
@@ -22,31 +32,87 @@ function addScriptToHead() { // eslint-disable-line
...
-      results.forEach((item) => {
-        // TODO: Create clickable divs for each item that is
-        // returned from the query and fill in input fields with appropriate
-        // information provided by the places object
-        console.log(item);
+      results.forEach((place) => {
+        div = document.createElement('div');
+        div.setAttribute('data-place-id', place.place_id);
+        div.innerText = place.name;
+        div.id = 'place';
```
megankj, line 66: use class instead of id here

src/main/webapp/destinationCreation.js
```diff
@@ -22,31 +32,87 @@ function addScriptToHead() { // eslint-disable-line
...
+
+  const placeService = new google.maps.places.PlacesService(map);
+  const request = {
+    placeId: place,
+    fields: ['name', 'geometry'],
+  };
+  let marker;
+  const element = document.querySelectorAll('#place');
+  let i;
+  for (i = 0; i < element.length; i++) {
```
megankj, line 94: same here

src/main/webapp/destinationCreation.js
```diff
@@ -22,31 +32,87 @@ function addScriptToHead() { // eslint-disable-line
...
+    removeElement(element[i].id);
+  }
+  placeService.getDetails(request, (result, status) => {
```
megankj, line 97: Do you need to make a second request here?  it seems like the first findPlaceFromQuery should return the geometry info.  Instead of making a second request to get this data again, you should pass this data to the onclick handler.

Tim might have a better way of doing this, but you can use [bind](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind) to bind the arguments to the method, like this: https://jsbin.com/midugowidi/edit?html,js,output

PR #60, opened 2020-08-19
Stores place id in datastore with the destination created

rysimone:


rysimone:
I created this branch off of this [one](https://github.com/googleinterns/step183-2020/pull/56) so that I could use the changes I made in this PR. It seems like doing so added my commits and files changed from the PR into this one.

PR #63, opened 2020-08-20
Add filtering for tags

shreyachatterjee00:
* In Destination.java, add functions for converting strings to Tag objects and finding the intersection between 2 sets using retainAll

* In GenerateServlet.java, add filtering for tags

* If user clicks no tags, no filtering is done 

* Comment out entire GenerateHuntTest.java, I'm going to write updated tests on that file for both this PR and the one for allowing the user to enter the number of Destinations they want

### DIFF
--- a/src/test/java/com/google/sps/GenerateHuntTest.java
+++ b/src/test/java/com/google/sps/GenerateHuntTest.java
@@ -1,6 +1,5 @@
 package com.google.sps.data;
-
-/*import com.google.sps.servlets.GenerateServlet;
+import com.google.sps.servlets.GenerateServlet;
 import java.util.Arrays;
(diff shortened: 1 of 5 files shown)


src/test/java/com/google/sps/GenerateHuntTest.java
```diff
@@ -13,111 +13,179 @@
 
 @RunWith(JUnit4.class)
 public final class GenerateHuntTest {
+  Set<Destination.Tag> goldenGateSet = new HashSet<>(Arrays.asList(Destination.Tag.HISTORICAL));
```
tjgoog, line 16: 👍 Good expansion of the test data

PR #66, opened 2020-08-24
Checks for duplicate destinations

rysimone:
After a user searches for a destination, results are presented from the user. When one of the results is clicked, datastore is checked for destinations with matching placeIds. If a duplicate is found, an alert message is displayed.


src/main/java/com/google/sps/servlets/DuplicateServlet.java
```diff
@@ -0,0 +1,76 @@
...
+  private static final Query QUERY = new Query(Constants.DESTINATION_ENTITY);
+  private static final PreparedQuery RESULTS = DATASTORE.prepare(QUERY);
+
+  private static final Gson GSON = new Gson();
+
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String placeId = request.getParameter(PLACE_ID_URL_PARAMETER);
+    ArrayList<Destination> destinations = getDestinationsFromDatastore();
+    if(searchForDuplicate(destinations, placeId)){
+      response.setContentType("text/html");
```
lilymzhou, line 48: I might move this line outside the if loop, since this is used in both the if and else clauses.

src/main/java/com/google/sps/servlets/DuplicateServlet.java
```diff
@@ -0,0 +1,75 @@
...
+  private static final String ORIGINAL = "original";
+  private static final String PLACE_ID_URL_PARAMETER = "place-id";
+
+  private static final DatastoreService DATASTORE = DatastoreServiceFactory.getDatastoreService();
+  private static final Query QUERY = new Query(Constants.DESTINATION_ENTITY);
+  private static final PreparedQuery RESULTS = DATASTORE.prepare(QUERY);
+
+  private static final Gson GSON = new Gson();
+
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
tjgoog, line 44: I think this should be GET rather than POST because it doesn't modify any state or data on the server side.

src/main/java/com/google/sps/servlets/DuplicateServlet.java
```diff
@@ -0,0 +1,75 @@
...
+    response.setContentType("text/html");
+    if (searchForDuplicate(destinations, placeId)) {
+      response.getWriter().println(DUPLICATE);
+    } else {
+      response.getWriter().println(ORIGINAL);
+    }
+  }
+
+  /* Query datastore for Destination objects, convert then to Destination class, store in ArrayList. */
+  public ArrayList<Destination> getDestinationsFromDatastore() {
```
tjgoog, line 56: Make this private too

src/main/webapp/destinationCreation.js
```diff
@@ -114,3 +116,17 @@ function removeElements(elemcls) {
     element[i].remove();
   }
 }
+
+/**
+ * Fetch the data from /duplicates servlet
```
tjgoog, line 121: Document what happens when a duplicate is found

src/main/java/com/google/sps/servlets/DuplicateServlet.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  /* Search for duplicate destinations based off of placeId*/
+  private boolean searchForDuplicate(ArrayList<Destination> destinations, String placeId) {
+    for (Destination destination : destinations) {
+      if (placeId.equals(destination.getPlaceId())) {
+        return true;
+      }
+    }
+    return false;
```
megankj, line 73: let me show you how this can be done with streams in one line with [anyMatch](https://www.geeksforgeeks.org/stream-anymatch-java-examples/)!
```suggestion
    destinations.stream().anyMatch(dest -> placeId.equals(dest.getPlaceId()));
```

rysimone: Wow..... streams are a beautiful thing lol

PR #67, opened 2020-08-25
Improve efficiency of entity extraction

lilymzhou:
- Add a loader that appears when the server begins to perform entity extraction, and disappears once the process is complete.
- Initially, check to see if the user's guess is a perfect match for the answer. If not, perform entity extraction.


src/main/webapp/go.css
```diff
@@ -131,3 +131,18 @@ button:hover a {
   font-size: 22px;
   font-weight: bold;
 }
+
+#submit-loader {
```
tjgoog, line 135: Does the loader need to be hidden by default on the initial page load?

lilymzhou: Yes -- the window.onload() function in go.js hides the loader when the page first loads.

PR #68, opened 2020-08-27
Dynamically load cities from datastore

shreyachatterjee00:
* Query the datastore to find cities that users have submitted
* Display those cities on generateHunt.html
* Add a test to ensure the cities returned are correct


src/test/java/com/google/sps/GenerateHuntTest.java
```diff
@@ -367,4 +390,42 @@ public void noDests() {
...
+    disneyEnt.setProperty(Constants.DESTINATION_JSON, disneyJson);
+    datastore.put(disneyEnt);
+
+    GetCitiesServlet cities = new GetCitiesServlet();
+    Set<String> actual = cities.getCities();
+
+    Set<String> expected = new HashSet<String>();
+    expected.add("San Francisco");
+    expected.add("Paris");
+    expected.add("Florida");
```
megankj, line 428: you might also want to have a destiantion with the same city to ensure it doesn't return dupes

PR #69, opened 2020-08-28
Tests for user input

rysimone:


src/test/java/com/google/sps/DestinationDataServletTest.java
```diff
@@ -0,0 +1,105 @@
...
+    helper.setUp();
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Test
+  public void storeInDatastore() throws IOException {
```
tjgoog, line 68: This isn't actually testing any of the behavior of DestinationDataServlet, which it should be, given the name.

src/main/java/com/google/sps/servlets/DestinationDataServlet.java
```diff
@@ -58,7 +58,7 @@
    * The destination object is then formatted into a JSON string and stored in datastore.
    */
   @Override
-  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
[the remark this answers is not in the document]
rysimone: Yea you are right. I fixed this in a different PR and forgot to do it here

src/test/java/com/google/sps/DestinationDataServletTest.java
rysimone: I test that the new entity is being stored and that is a behavior of destination data servlet. I am trying to see how to mock html messages to test if no name is passed and other required fields.

megankj: Your test shouldn't be testing the same _behavior_ as DestinationDataServlet, it should test the actual method inside of DestinationDataServlet, since that's the code we want to be sure is behaving as expected.

rysimone: Ok. I tried calling the doPost method inside of DestinationDataServlet to check that what is being created is accurate, but I was getting api errors so I'm not sure how I am supposed to test that. Should I write tests for my convertTagsToEnums and convertObscurityToEnum mehods??

tjgoog: Could you share the API errors? Maybe the test version of the datastore isn't being loaded by the servlet, or the mocks are missing some setup. Seeing the errors will help debug.

rysimone:
Those error messsages were a result of me using mockito to mock datastore, HTTP request, and response. Then calling servlet.doPost(response, request). The error arose when line 109 `Entity destinationEntity = new Entity(Constants.DESTINATION_ENTITY);` was called. I have since changed my implementation so I do not have the exact error message.

src/main/webapp/destinationCreation.js
```diff
@@ -123,7 +123,7 @@ function removeElements(elemcls) {
  * @param {String} data placeId to be checked if it is already in datastore
  */
 function searchForDuplicate(data) {
-  fetch('/duplicates?' + PLACE_ID_URL_PARAM + '=' + data)
+  fetch('/duplicates?' + PLACE_ID_URL_PARAM + '=' + data, {method: 'POST'})
```
megankj, line 126: this doesn't belong in this pull request, right?

tjgoog (approved):

src/main/webapp/destinationCreation.js
megankj: this shouldn't be POST, right?

src/test/java/com/google/sps/DestinationDataServletTest.java
```diff
@@ -0,0 +1,90 @@
...
+import org.mockito.MockitoAnnotations;
+
+@RunWith(JUnit4.class)
+public final class DestinationDataServletTest {
+  @Mock private HttpServletRequest request;
+  @Mock private HttpServletResponse response;
+
+  private DestinationDataServlet servlet;
+
+  private static final String NAME_PARAMETER = "name";
```
megankj, line 34: Instead of redefining these constants, you could make the corresponding fields public in DestinationDataServlet and then just use them like  DestinationDataServlet.NAME_PARAMETER

src/test/java/com/google/sps/DestinationDataServletTest.java
```diff
@@ -0,0 +1,90 @@
...
+    doReturn("Golden-red in color").when(request).getParameter(HINT2_PARAMETER);
+    doReturn("You have to pay to use me").when(request).getParameter(HINT3_PARAMETER);
+    String[] returnedTags = {"historical", "tourist"};
+    doReturn(returnedTags).when(request).getParameterValues(TAG_PARAMETER);
+    doReturn("easy").when(request).getParameter(OBSCURITY_PARAMETER);
+    doReturn("123").when(request).getParameter(PLACEID_PARAMETER);
+    servlet.doPost(request, response);
+
+    Assert.assertEquals(
+        1, datastore.prepare(new Query(Constants.DESTINATION_ENTITY)).countEntities(withLimit(10)));
```
megankj, line 86: You should probably also verify that the one entry in datastore is what you expect it to be.  Also do you need to do the withLimit part?  It doesn't really matter because you are just checking to see that the size is one, but this seems unnecessary/irrelevant to the test.

megankj (changes requested):

src/test/java/com/google/sps/DestinationDataServletTest.java
rysimone: I just tried to run it without the withLimit part and it works so I do not need it and will remove it.

src/test/java/com/google/sps/DestinationDataServletTest.java
```diff
@@ -0,0 +1,148 @@
...
+    doReturn(PLACEID_INPUT).when(request).getParameter(PLACEID_PARAMETER);
+    servlet.doPost(request, response);
+
+    Assert.assertEquals(
+        1, datastore.prepare(new Query(Constants.DESTINATION_ENTITY)).countEntities());
+
+    Query query = new Query(Constants.DESTINATION_ENTITY);
+    PreparedQuery results = datastore.prepare(query);
+
+    ArrayList<Destination> destinations = new ArrayList();
```
tjgoog, line 107: Because this will always have a single destination entity, you should be able to just use `results.asSingleEntity()` instead of building a list.

tjgoog (changes requested):

tjgoog (approved):

src/test/java/com/google/sps/DestinationDataServletTest.java
```diff
@@ -0,0 +1,135 @@
...
+            (String) results.asSingleEntity().getProperty(Constants.DESTINATION_JSON),
+            Destination.class);
+
+    String actual = GSON.toJson(destination);
+
+    String expected = getExpectedDestination();
+
+    Assert.assertEquals(actual, expected);
+
+    verify(response).sendRedirect(DestinationDataServlet.HOME_URL);
```
megankj, line 104: Tests should be structured in three distinct chunks: 1) arrange 2) act 3) assert.  This gives the reader a visual clue about what the test code is doing.

Therefore, you should rearrange your code like this:
// Arrange
DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
servlet = new DestinationDataServlet(datastore);
doReturn....

// Act
servlet.doPost(request, response);

// Assert
PreparedQuery results = ...
String actual = ...
Assert.assertEquals(1, results.countEntities());
Assert.assertEquals(actual, getExpectedDestination());

megankj (approved):

PR #72, opened 2020-08-31
Use Place IDs to compare user's guess with the correct destination name.

lilymzhou:
Demo:
https://lilymzhou-step-2020.uc.r.appspot.com/

### DIFF
--- a/src/main/webapp/go.js
+++ b/src/main/webapp/go.js
@@ -286,10 +286,7 @@ function getAutoHint() {
  * destination, before getting a new hint.
  */
 function generatePlaceID() {
-  const request = {
-    query: hunt.getCurDestName(),
-    fields: ['place_id'],
-  };
+  const request = createRequestForPlaceID(hunt.getCurDestName());
   service.findPlaceFromQuery(request, function(results, status) {
     if (status === google.maps.places.PlacesServiceStatus.OK) {
       hunt.setPlaceID(results[0].place_id);
@@ -298,6 +295,17 @@ function generatePlaceID() {
       updateMessage(GENERATE_DISPLAY, NO_PLACEID_MSSG);
     }
   });
+}
+
+/**
+ * @param {String} queryName Query for place search
+ * @return {String} request to be used in place search
+ */
+function createRequestForPlaceID(queryName) {
+  return request = {
+    query: queryName,
+    fields: ['place_id'],
+  };
 }
 
 /**
@@ -437,10 +445,7 @@ function checkUserDestinationGuess() { //eslint-disable-line
  * @param {String} userGuess User's guess for current destination.
  */
 function checkGuessWithPlaceIDs(userGuess) {
-  const answerRequest = {
-    query: hunt.getCurDestName(),
-    fields: ['place_id'],
-  };
+  const answerRequest = createRequestForPlaceID(hunt.getCurDestName());
   service.findPlaceFromQuery(answerRequest, function(results, status) {
     if (status === google.maps.places.PlacesServiceStatus.OK) {
       const answerID = results[0].place_id;
@@ -456,10 +461,7 @@ function checkGuessWithPlaceIDs(userGuess) {
  * destination
  */
 function getPlaceIDOfGuess(userGuess, answerID) {
-  const userRequest = {
-    query: userGuess,
-    fields: ['place_id'],
-  };
+  const userRequest = createRequestForPlaceID(userGuess);
   service.findPlaceFromQuery(userRequest, function(results, status) {
     if (status === google.maps.places.PlacesServiceStatus.OK) {
       const guessID = results[0].place_id;
--- a/src/main/webapp/go.js
+++ b/src/main/webapp/go.js
@@ -428,12 +428,7 @@ function checkUserDestinationGuess() { //eslint-disable-line
 
   // Check to see if userGuess can be used to identify the correct place
   // using the Places library.
-  const answerID = getPlaceIDOfAnswer();
-  const guessID = getPlaceIDOfGuess(userGuess);
-  if (answerID != '' && guessID != '' && guessID === answerID) {
-    handleDestinationAnswer(/* correct = */ true);
-    return;
-  }
+  checkUserGuessWithPlaceID(userGuess);
 
   // Perform entity extraction.
   const queryStr = GUESS_URL + '?' + GUESS_INPUT + '=' + userGuess +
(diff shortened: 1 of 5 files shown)


src/main/webapp/go.js
```diff
@@ -436,6 +445,44 @@ function checkUserDestinationGuess() { //eslint-disable-line
...
+/**
+ * Retrieves the place ID of the current destination, retrieved
+ * using the Places library.
+ */
+function getPlaceIDOfAnswer() {
+  const answerRequest = {
+    query: hunt.getCurDestName(),
+    fields: ['place_id'],
+  };
+  service.findPlaceFromQuery(answerRequest, function(results, status) {
```
megankj, line 457: this is an async function so you should await the result

src/main/webapp/go.js
```diff
@@ -436,6 +445,44 @@ function checkUserDestinationGuess() { //eslint-disable-line
...
+    query: userGuess,
+    fields: ['place_id'],
+  };
+  service.findPlaceFromQuery(userRequest, function(results, status) {
+    if (status === google.maps.places.PlacesServiceStatus.OK) {
+      return results[0].place_id;
+    } else {
+      return '';
+    }
+  });
```
megankj, line 483: this is a lot of duplicated code - create a helper method for this

src/main/webapp/go.js
```diff
@@ -436,6 +445,44 @@ function checkUserDestinationGuess() { //eslint-disable-line
   });
 }
 
+/**
+ * Retrieves the place ID of the current destination, retrieved
+ * using the Places library.
+ */
+function getPlaceIDOfAnswer() {
```
tjgoog, line 452: There's a lot of overlap between this function and `generatePlaceId()`. Could you factor out the common part to avoid code duplication?

src/main/webapp/go.js
```diff
@@ -436,6 +445,44 @@ function checkUserDestinationGuess() { //eslint-disable-line
...
+  });
+}
+
+/**
+ * Retrieves the place ID of the user's guess, retrieved
+ * using the Places library.
+ * @param {String} userGuess The user's guess for the name of the
+ * current destination.
+ */
+function getPlaceIDOfGuess(userGuess) {
```
tjgoog, line 472: I think this should also be changed to use a single implementation of the common code referred to in my comment above.

src/main/webapp/go.js
tjgoog: This function doesn't return a Promise though, so the current callback function structure seems the most natural to me.

megankj: In that case I think the callback make the second call to findPlaceFromQuery.  I'm thinking that since these APIs have to make network requests to get the result that you should make sure you have values to evaluate.

src/main/webapp/go.js
lilymzhou: @megankj 

I decided to nest the second call to findPlaceFromQuery inside the first call -- I put some of the duplicated code into a common function (createRequestForPlaceID), but I'm not sure how to combine the rest of the redundant code.

src/main/webapp/go.js
```diff
@@ -426,7 +434,54 @@ function checkUserDestinationGuess() { //eslint-disable-line
     return;
   }
 
-  // If userGuess is not an exact match, perform entity extraction.
+  // Check to see if userGuess can be used to identify the correct place
+  // using the Places library.
+  checkGuessWithPlaceIDs(userGuess);
```
megankj, line 439: This method is doing more than just checking the guess with place ID.  If the place ID doesn't match, it will also use entity extraction.

src/main/webapp/go.js
```diff
@@ -426,7 +434,54 @@ function checkUserDestinationGuess() { //eslint-disable-line
...
+}
+
+/**
+ * Retrieves place ID for the answer, and compares this to the place
+ * ID retrieved for the user's guess.
+ * @param {String} userGuess User's guess for current destination.
+ */
+function checkGuessWithPlaceIDs(userGuess) {
+  const answerRequest = createRequestForPlaceID(hunt.getCurDestName());
```
megankj, line 448: Place IDs for each destination are stored when the Destination object is created.  Instead of making the request to get it here, it would be simpler to just propagate the place ID from the Destination object to the HuntItem object.  Then you don't have to make this request because you already have the place ID.

src/main/webapp/go.js
```diff
@@ -426,7 +434,54 @@ function checkUserDestinationGuess() { //eslint-disable-line
...
+  });
+}
+
+/**
+ * @param {String} userGuess The user's guess for the name of the
+ * current destination.
+ * @param {String} answerID Place ID corresponding to the correct
+ * destination
+ */
+function getPlaceIDOfGuess(userGuess, answerID) {
```
megankj, line 463: This method isn't just getting the Place ID of the guess.  If it doesn't match it falls back to entity extraction.

src/main/webapp/go.js
tjgoog: +1, but it's an optional suggestion as far as I'm concerned. Megan may feel more strongly.

I originally hesitated to suggest this because HuntItems currently don't contain the place ID, but thinking about it further, it shouldn't be too much work to start storing place IDs in them, and you don't have to worry about being backward compatible with old hunts in the datastore because people are always generating new hunts instead of using an existing ID.

src/main/webapp/go.js
```diff
@@ -286,10 +286,7 @@ function getAutoHint() {
  * destination, before getting a new hint.
  */
 function generatePlaceID() {
-  const request = {
-    query: hunt.getCurDestName(),
-    fields: ['place_id'],
-  };
+  const request = createRequestForPlaceID(hunt.getCurDestName());
```
megankj, line 289: do you need this function anymore now that each place already has a placeID set at construction?

src/main/webapp/go.js
```diff
@@ -426,7 +434,44 @@ function checkUserDestinationGuess() { //eslint-disable-line
     return;
   }
 
-  // If userGuess is not an exact match, perform entity extraction.
+  // Check to see if userGuess can be used to identify the correct place
+  // using the Places library.
+  // checkGuessWithPlaceIDs(userGuess);
```
megankj, line 439: remove this comment?

src/main/webapp/go.js
```diff
@@ -436,31 +406,20 @@ function checkUserDestinationGuess() { //eslint-disable-line
...
  * @param {String} userGuess The user's guess for the name of the
  * current destination.
  * @param {String} answerID Place ID corresponding to the correct
  * destination
  */
-function getPlaceIDOfGuess(userGuess, answerID) {
+function checkGuessWithIDOrEntities(userGuess, answerID) {
   const userRequest = createRequestForPlaceID(userGuess);
   service.findPlaceFromQuery(userRequest, function(results, status) {
     if (status === google.maps.places.PlacesServiceStatus.OK) {
```
tjgoog, line 425: To reduce redundant code (`checkGuessWithEntities(userGuess)`), I suggest making this line
```suggestion
    if (status === google.maps.places.PlacesServiceStatus.OK && results[0].place_id === answerID) {
```
Then you can simplify down to a single `if` branch and single `else` branch.

tjgoog (approved):

megankj (approved):

PR #75, opened 2020-09-01
Creates a success page for after a destination is created

rysimone:


rysimone:
image

src/main/webapp/destinationCreation.js
```diff
@@ -123,7 +123,7 @@ function removeElements(elemcls) {
  * @param {String} data placeId to be checked if it is already in datastore
  */
 function searchForDuplicate(data) {
-  fetch('/duplicates?' + PLACE_ID_URL_PARAM + '=' + data)
+  fetch('/duplicates?' + PLACE_ID_URL_PARAM + '=' + data, {method: 'POST'})
```
tjgoog, line 126: Shouldn't this still be GET? GET is the default.

src/main/webapp/destinationCreationSuccess.html
```diff
@@ -0,0 +1,23 @@
...
+    <script src="success.js"></script>
+  </head>
+  <body class="background" onload="confirmCreation()">
+    <div class="content">
+      <div id="confirmation-block">
+        <h1>Destination Successfully Created!</h1>
+      </div>
+      <div class="button-row">
+        <a href="index.html" class="button">Home Page</a>
+        <a href="generateHunt.html" class="button">Generate a Hunt</a>
```
megankj, line 19: maybe you should have another button to add another destination?

rysimone: Yea that's not a bad idea

